PR #25, opened 2023-12-04
Add 100x100 pixel logo to README

### LINKED ISSUE #23: Add logo to repository
I don't think we can replace the NCAR|UCAR logo that appears next to `CUPiD` (that is inherited from the NCAR organization), but we could add it to the README like https://github.com/NCAR/DART does

mnlevy1981:
Currently on the same line as the top header, but it may look better to put the logo on a line by itself?

Fixes #23


mnlevy1981:
To preview the README, click [here](https://github.com/NCAR/CUPiD/blob/ac900a09c98bf3ef7da567d1187ff89d27787efe/README.md); also including a screenshot:

README with logo

mnlevy1981:
The only other thought I had was to create a top-level `docs/` directory and move `logo.png` to `docs/images/logo.png`, but I think it makes more sense to wait and see where we want the documentation to live / how we want it structured and then move the logo and update the source of the logo in the README file

TeaganKing:
Are there other documentation structures that you think make more sense than a `docs` directory (and a wiki page)? I think we will probably be quickly adding more documentation, so it actually may not be a bad idea to have this directory structure set up already?

mnlevy1981:
I was picturing a `docs/` directory that uses `sphinx` to build the full CUPiD documentation, which I don't think would be trivial to spin up. A simple webpage might be more appropriate, but in either case we probably want more code in the repository before we determine the best way to document it.

TeaganKing:
Ok, sounds good! Yeah, `sphinx` can take a bit of work to set up...

PR #26, opened 2023-12-06
Update README.md to include build / run steps

mnlevy1981:


mnlevy1981:
I made this a draft PR because we should test it out and make sure everything still works. @rmshkv and I made a few changes on the ADF side (and possibly in `examples/adf-mom6`, so there might be another PR coming?); there are still a couple of random `scratch` directories where the data hasn't been copied to `campaign` yet, but for now I think ADF will be happy reading the existing data and won't try to write anything anywhere it has read-only permissions.

README.md
```diff
@@ -2,9 +2,50 @@
...
+```
+
+Note that `conda` now defaults to using `mamba` to solve environments;
+the `mom6-environment.yml` environment is complicated,
+so older versions of `conda` should be updated (`conda update -n base conda`) or you should use `mamba` instead.
+
+## Running
+
+CUPiD currently provides two examples for generating diagnostics.
+To test the package out, try to run `examples/adf-mom6`:
```
TeaganKing, line 44: Do we want to recommend doing this from a jupyterhub terminal for ease of viewing the notebook's?

mnlevy1981: We definitely need a comment about how to view the finished product, but I'm hesitant to recommend running everything through the Hub. Maybe a comment like

> You can use Jupyter to view generated notebooks in `${CUPID_ROOT}/path/to/notebooks` or you can copy the entire `${CUPID_ROOT}/path/to/HTML` directory to your local machine and look at the generated HTML in a web browser

TeaganKing: Yes, I'm supportive of that!

PR #28, opened 2023-12-07
Update quickstart instructions

TeaganKing:
A few minor updates to make sure that quick start instructions run smoothly (I did get everything successfully set up with these steps).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,15 @@ Then build the necessary conda environments with
 
 ```
 $ conda env create -f nbscuid/dev-environment.yml
+$ which nbscuid-run
 $ conda env create -f mom6-environment.yml
+```
+
+IF `which nbscuid-run` returned the error `which: no nbscuid-run in ($PATH)`, then please run the following:
+
+```
+$ conda activate nbscuid-dev
+$ pip install -e .  # installs nbscuid
 ```
 
 Note that `conda` now defaults to using `mamba` to solve environments;
@@ -45,7 +53,6 @@ To test the package out, try to run `examples/adf-mom6`:
 
 ```
 $ conda activate nbscuid-dev
-$ pip install -e .  # installs nbscuid
 $ cd examples/adf-mom6
 $ nbscuid-run config.yml
 $ nbscuid-build config.yml # Will build HTML from Jupyter Book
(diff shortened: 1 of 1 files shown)


mnlevy1981:
Before we update the instructions, the [`dev-environment.yml`](https://github.com/NCAR/CUPiD/blob/main/nbscuid/dev-environment.yml#L5-L6) file contains

```yaml
    - pip:
      - -e ../
```

which should have installed `nbscuid` for you. I could see four possibilities:

1. the `-e ../` is a newer construct and your version of `conda` is too old to support it
2. the `-e ../` is an older constrcut that has been deprecated and your version of `conda` is too new to support it
3. Something else went wrong in the install (if I remember right, it looked like it installed cleanly but you were subsequently having disk quota issues) and it wasn't clear that that `conda env create` command didn't complete successfully
4. one of the first two, but there is an alternate way to include this in the environment file that is supported by the current release version of `conda` as well as enough old versions that we can safely assume that if it doesn't work it's okay to request users update to something newer

One possible quick work-around would be to add

```diff
$ conda env create -f nbscuid/dev-environment.yml
+ $ which nbscuid-run
$ conda env create -f mom6-environment.yml
```

and have users run `pip install -e .` if the `which` command returns `which: no nbscuid-run in ($PATH)` error. Hopefully for most users the `conda env create` will install the package correctly and then `nbscuid-run` will be available.

TeaganKing:
Good point, @mnlevy1981 !
I could see 1 or 2 being possibilities. That is correct that I had disk quota errors, but I ran the environment setup from scratch after that happened, so I'm assuming 3 probably wasn't the cause of the error. I updated the instructions to use which nbscuid-run.

mnlevy1981:
This looks great! One more small suggestion -- I just looked at [the entire README](https://github.com/NCAR/CUPiD/blob/4cdb53d38be034428ecc6b0d79275a252e458a8c/README.md), and noticed that the end of the "run" section is

> Note that `conda` now defaults to using mamba to solve environments; the `mom6-environment.yml` environment is complicated, so older versions of `conda` should be updated (`conda update -n base conda`) or you should use `mamba` instead.

Should we change this to something like

> Notes:
> 
> 1. `conda` now defaults to using mamba to solve environments; the `mom6-environment.yml` environment is complicated, so older versions of `conda` should be updated (`conda update -n base conda`) or you should use `mamba` instead.
> 2. IF `which nbscuid-run` returned the error `which: no nbscuid-run in ($PATH)`, then please run the following:
> 
>     ```
>     $ conda activate nbscuid-dev
>     $ pip install -e .  # installs nbscuid
>     ```

mnlevy1981 (approved):
looks great!

PR #32, opened 2023-12-12
Reorganize externals

mnlevy1981:
The main purpose of this PR is to commit a copy `nblibrary` from https://github.com/rmshkv/nbscuid-adf-sandbox.git into CUPiD rather than bringing it in as an external. We now use `manage_externals` to bring in ADF code, and also to bring in MOM6 code (needed for the `mom6-tools-nbscuid` conda environment).

Aside from creating the `nblibrary` directory, most of the updates are small (changing paths to point to the new locations of `mom6-tools` and `ADF`; updating the README, etc). @rmshkv after accepting this PR I don't think you want to do any more development in `nbscuid-adf-sandbox` -- that should all happen in this repo now.


mnlevy1981:
Forgot to mention that I did some testing... With these changes I can still run `nbscuid-run config.yml` from `examples/adf-mom6` and `nbscuid-run config-adf-only.yml` from `examples/adf-only`

README.md
```diff
@@ -18,7 +18,7 @@ CUPiD is a collaborative effort that unifies all CESM component diagnostics and
 To install CUPiD, you need to check out the code and then set up a few environments.
 The initial examples have hard-coded paths that require you to be on `casper`.
 
-The code relies on submodules to install `manage_externals` and then uses `manage_externals` for one more package, so the `git clone` process is a little more complicated than usual:
+The code relies on submodules to install `manage_externals` and then uses `manage_externals` for two more packages, so the `git clone` process is a little more complicated than usual:
```
TeaganKing, line 21: If `manage_externals` is going to continue to bring in additional packages, should we specify the types of packages it is bringing in instead of the number of packages? Eg, component-specific diagnostic packages & NBscuid 'framework'? Then this would be more helpful for people adding to the code, as well as reduce the need to update the README regularly.

README.md
```diff
@@ -30,15 +30,17 @@ Then build the necessary conda environments with
 
 ```
 $ conda env create -f nbscuid/dev-environment.yml
+$ conda activate nbscuid-dev
```
TeaganKing, line 33: Good catch, thanks!

README.md
```diff
@@ -30,15 +30,17 @@ Then build the necessary conda environments with
...
 $ which nbscuid-run
+$ conda deactivate
```
TeaganKing, line 35: Maybe this helps for clarity, but I think you can also just create a new environment while still in nbscuid-dev? Maybe one less step would be easier?

rmshkv (approved):
Didn't have a chance to look super closely, but bringing in nblibrary directly sounds great to me!

mom6-environment.yml
```diff
@@ -23,4 +23,4 @@ dependencies:
   - yaml
   - zarr
   - pip:
-    - git+https://github.com/NCAR/mom6-tools.git
+    - -e externals/mom6-tools
```
TeaganKing, line 26: Just to clarify, since this is brought in as an external, it will still get updated any time the repo is updated, right?

One other slightly bigger-picture thing-- should folks be adding tests in the external repositories to make sure they don't break in CUPiD if they are updated?

README.md
mnlevy1981: Good point! I think I cleared it up -- `manage_externals` is a temporary fix so we can help develop outside packages (`ADF` and `mom6-tools` currently, perhaps `geocat.comp` and / or `.viz` in the near future); at some point we can hopefully rely on a conda environment with tagged versions instead.

README.md
mnlevy1981: I was on the fence about adding this line in; I think you're right, and it's easier to leave it out

mom6-environment.yml
TeaganKing: Ok, that sounds like a good idea. I guess I'm just also thinking that it would be good to test things before breaking CUPiD (eg, before code is merged into main in the external repositories)

TeaganKing (approved):
I was able to run through the environment creation and basic example after also running `git submodule update --init` in order to check out externals (which probably won't need to be done after this is merged in). So, if the other comments are addressed, this PR looks fine to me!

Issue #38, opened 2024-01-09
Update documentation


TeaganKing:
Some documentation currently lives here: https://nbscuid.readthedocs.io/en/latest/

It'd be great to update this to reflect newer code structure.

TeaganKing:
The documentation currently lives here: https://ncar.github.io/CUPiD/index.html

We should also figure out how to directly add the readme file to `index.rst` (rather than just include hyperlink), as @rmshkv suggested!

anissa111:
> The documentation currently lives here: [ncar.github.io/CUPiD/index.html](https://ncar.github.io/CUPiD/index.html)
> 
> We should also figure out how to directly add the readme file to `index.rst` (rather than just include hyperlink), as @rmshkv suggested!

I can take this on!

TeaganKing:
The documentation for setting up a new notebook that is compatible with CUPiD is now complete, and the readme is now on our documentation page, as well, as mentioned in comments above.

The next item to do would be to focus on including specific function documentation with `sphinx` and `readthedocs`.

TeaganKing:
Additionally, we should add an 'adding scripts' page to the documentation

mnlevy1981:
Replaced with #73 and #74

PR #39, opened 2024-01-09
Drop nbscuid name

mnlevy1981:
I renamed `nbscuid/` -> `cupid/`, and tried to replace every instance of nbscuid with cupid (including the conda environment names).
I've also updated the README file and verified the instructions provided still work -- this required changing a few paths in the nblibrary YAML file because directories on scratch had been scrubbed.


pyproject.toml
```diff
@@ -36,10 +38,10 @@ dependencies = [
 ]
 
 [project.urls]
-repository = "https://github.com/rmshkv/nbscuid"
+repository = "https://github.com/NCAR/CUPiD"
 documentation = "https://nbscuid.readthedocs.io"
```
TeaganKing, line 42: This should be updated in #38

README.md
```diff
@@ -30,21 +30,21 @@ $ ./manage_externals/checkout_externals
...
 
 ```
-$ conda env create -f nbscuid/dev-environment.yml
-$ conda activate nbscuid-dev
-$ which nbscuid-run
-$ conda env create -f mom6-environment.yml
+$ conda env create -f cupid/dev-environment.yml
+$ conda activate cupid-dev
+$ which cupid-run
+$ conda env create -f cupid-analysis.yml
```
TeaganKing, line 36: This is a clear new environment name, thanks for replacing env with analysis, too.

rmshkv (approved):
Looks good to me, very thorough! :)

README.md
mnlevy1981: Yeah, I figured if we were moving to a single environment for all diagnostics we should have a generic name :) Seeing this block of text, though, I wonder if the file name should be `analysis-environment.yml`

TeaganKing: I feel like -environment is redundant? And then at least if people are using the environment on a machine with lots of additional environments it may help to know that it's the CUPiD one?

TeaganKing:
I also double checked all of the nbscuid places that I found against this list of changes, and it looks great to me!

TeaganKing (approved):

README.md
mnlevy1981: The environment could stay `cupid-analysis` if the filename changed, and then it would match `dev-environment.yml`. I'll start a new issue ticket, though, because I think we also want an `envs/` directory that contains both environment files rather than having one in `cupid/` and the other in the top level

TeaganKing: Ok, seems reasonable to me! Thanks Mike!

rmshkv: One more thought, actually - I feel like I've seen the standard for environment spec YAML files be to always call the file "environment.yml", and I specifically made the choice to add a descriptor to these ones for clarity, but given that we also have various config YAML files floating around between the different component diagnostics and CUPiD usecases, keeping "environment" might actually be good to lessen confusion.

rmshkv: (For additional context - the name of the environment is specified in the file and isn't linked to the yml file name)

PR #41, opened 2024-01-11
Support Derecho as well as Casper

mnlevy1981:
The main fix on this branch is moving more data from `/glade/cheyenne/scratch` and `/glade/derecho/scratch` to `/glade/campaign` (some intermediate ADF computations from @rmshkv's scratch space as well as output from an F compset from @cecilehannay's scratch space). I also cleared the output from the notebook templates in `examples/nblibrary`, and removed some trailing white space from `run.py` and `config.yml`.


mnlevy1981:
Testing: I ran the `adf-mom6` example on both machines successfully. Also, `git grep scratch` no longer finds any instances of that directory.

TeaganKing:
This looks good to me! The only other note I wanted to add is that @mnlevy1981 and I discussed future YAML configurations to point to users' own run and archive directories so that they can run in scratch. For CUPiD development, we don't want to have CESM output scrubbed unexpectedly.

TeaganKing (approved):

PR #49, opened 2024-01-18
point to shared directory, no dask

wwieder:
hopefully improves availability for users

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,9 @@ $ cupid-build config.yml # Will build HTML from Jupyter Book
 After the last step is finished, you can use Jupyter to view generated notebooks in `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run`
 or you can copy the entire `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run/_build/html`
 directory to your local machine and look at `index.html` in a web browser.
+
+The easiest way to visualize the web page on a browser is to:
+1. open a new browser window that points to https://fastx.ucar.edu:3300/session/`
+2. open a default desktop icon
+3. select the browser client 
+4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`, assuming your CUPiD is in your home directory 
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ After the last step is finished, you can use Jupyter to view generated notebooks
 or you can copy the entire `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run/_build/html`
 directory to your local machine and look at `index.html` in a web browser.
 
-The easiest way to visualize the web page on a browser is to:
+For users running on the NCAR super computers (derecho or casper), you can visualize the web page in a browser with the following steps:
 1. open a new browser window that points to https://fastx.ucar.edu:3300/session/`
 2. open a default desktop icon
 3. select the browser client 
-4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`, assuming your CUPiD is in your home directory 
+4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`
--- a/README.md
+++ b/README.md
@@ -64,7 +64,9 @@ or you can copy the entire `${CUPID_ROOT}/examples/coupled-model/computed_notebo
 directory to your local machine and look at `index.html` in a web browser.
 
 For users running on the NCAR super computers (derecho or casper), you can visualize the web page in a browser with the following steps:
-1. open a new browser window that points to https://fastx.ucar.edu:3300/session/`
-2. open a default desktop icon
-3. select the browser client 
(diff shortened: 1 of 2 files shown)


examples/nblibrary/land_comparision.ipynb
```diff
@@ -65,11 +65,9 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "e0d67916-85c5-4590-b03f-c00753e88df4",
+   "cell_type": "raw",
+   "id": "2fd07a40-172e-435e-84e9-e422ae920220",
```
mnlevy1981, line 69: If it works without these cells, can we delete them instead of leaving them raw?

examples/nblibrary/land_comparision.ipynb
```diff
@@ -102,11 +100,9 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "590d68d1-7bf3-4164-911c-c6593d010246",
+   "cell_type": "raw",
+   "id": "e6c2ec6c-628c-4b0a-bb00-a2739422b4d8",
```
mnlevy1981, line 104: Same as above... if it works without these cells, can we delete them instead of leaving them raw?

mnlevy1981 (changes requested):
A few minor suggestions, but overall I think this is great -- the land notebook ran for me while the daskified version hung due to the hard-coded project key.

README.md
```diff
@@ -62,3 +62,9 @@ $ cupid-build config.yml # Will build HTML from Jupyter Book
 After the last step is finished, you can use Jupyter to view generated notebooks in `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run`
 or you can copy the entire `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run/_build/html`
 directory to your local machine and look at `index.html` in a web browser.
+
+The easiest way to visualize the web page on a browser is to:
+1. open a new browser window that points to https://fastx.ucar.edu:3300/session/`
+2. open a default desktop icon
+3. select the browser client 
+4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`, assuming your CUPiD is in your home directory
```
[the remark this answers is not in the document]
wwieder: I'm also not sure where this should go, but  I was trying to remove frustrations of new users wondering what they had done after building their webpage.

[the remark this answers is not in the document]
wwieder: See if changes to readme reflect your ideas, @mnlevy1981 .

mnlevy1981:
Also, the descriptive text at the top of the notebook says "calcultes 20 year mean" -- can you change this to "calculates 5 year mean"? (add the "a" to calculates, and change the 20 to 5)

README.md
```diff
@@ -63,8 +63,8 @@ After the last step is finished, you can use Jupyter to view generated notebooks
 or you can copy the entire `${CUPID_ROOT}/examples/coupled-model/computed_notebooks/quick-run/_build/html`
 directory to your local machine and look at `index.html` in a web browser.
 
-The easiest way to visualize the web page on a browser is to:
+For users running on the NCAR super computers (derecho or casper), you can visualize the web page in a browser with the following steps:
 1. open a new browser window that points to https://fastx.ucar.edu:3300/session/`
 2. open a default desktop icon
 3. select the browser client 
-4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`, assuming your CUPiD is in your home directory 
+4. type `firefox CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`
```
mnlevy1981: I was playing with this, and it looks like when you just start typing to run a command, (a) you can't copy / paste from your local machine's clipboard, and (b) it's really easy to lose the command if you switch tabs to see what you're supposed to type. It might be easier to replace this last step with

4. type `xterm` and hit enter to launch a terminal window
5. in the terminal, run `cd /path/to/CUPiD/examples/coupled_model/computed_notebooks/quick-run/_build/html/index.html`
6. run `firefox index.html &`

mnlevy1981: sorry, first pass included `index.html` in the `cd` command, which is not a good idea

wwieder: Do you want to make this change?

README.md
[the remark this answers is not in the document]
mnlevy1981: I think I was typing in a different tab, but I did have one more suggestion below (having users launch a terminal and then run `firefox` from the terminal will probably be less frustrating since they don't need to type in a long path in a temporary window).

README.md
mnlevy1981: yeah, I think replacing [4] with [4], [5], [6] above is good; the second comment was just because I needed to edit the first comment due to a typo in [5]

mnlevy1981:
@wwieder it looks like I can push changes to your branch -- do you want me to add one more commit with my changes to the README and removing the raw cells from the notebook? Also, I noticed you have a section header

> Big blob of passive C with depth in pAD

 With two empty cells below it. Are you planning to add a few more computations, or can we remove this section / the empty cells?

wwieder:
yes, you can commit the readme changes. it's also fine to reduce that
header at the end of the notebook.

PR #50, opened 2024-01-18
Add a seaice example

dabail10:
This is from the tutorial, but now uses the single-variable timeseries files from the ADF. Note that I could not get the unweighted mean thing working. Right now it plots ice area (aice) from two cases and the difference. Then it computes the total ice area for the NH from one case and plots it.


mnlevy1981:
Thanks for starting this, @dabail10! I'd like to get #55 merged before this PR -- it has some updates to `config.yml` that I think we want before we update that file on this branch to make sure ploomber calls the new `seaice.ipynb` notebook.

Issue #44, opened 2024-01-18
More details in README.md "Installing" section would be helpful


erogluorhan:
Under the "Notes" part (1), you may want to detail since what conda version (23.10.0), conda started to use mamba as package solver. Maybe in addition, how to check conda version to make sure about this

PR #52, opened 2024-01-19
Update conda config

mnlevy1981:
Moved environment files to a common directory, and also added CUPiD dependencies to `dev-environment.yml` so they get installed by conda instead of pip


mnlevy1981:
I think I need to update `README.md` a little bit more to fully address #44

rmshkv (approved):

PR #55, opened 2024-01-22
Pass paths via config

mnlevy1981:
Use `config.yml` to pass configuration details (such as the directory containing CESM output) to both `land_comparison.ipynb` and `ocean_surface.ipynb`. Note that both of those are new names for old notebooks -- the land notebook had a typo in the name (comparison was misspelled) and the ocean notebook had the more generic name `surface.ipynb`


mnlevy1981:
I removed the dask client from the ocean notebook, and the land & atmosphere notebooks were both already serial. On derecho, it took ~11 minutes to run:

```
(cupid-dev) $ time cupid-run config.yml
real	11m2.020s
user	14m50.714s
sys	0m28.384s
(cupid-dev) $
```

mnlevy1981:
729064f switches to the serial task manager, which lets users follow progress on stdout. On derecho, it took a little longer than the parallel manager (11:30 instead of 11:02), which I think is a worthwhile trade-off:

```
real	11m35.800s
user	4m16.428s
sys	0m22.316s
```

examples/coupled_model/config.yml
```diff
@@ -68,16 +68,32 @@ compute_notebooks:
...
+          oce_cat: /glade/u/home/gmarques/libs/oce-catalogs/reference-datasets.yml
+
 
 
-  land_comparision:
+  land_comparison:
     parameter_groups:
       none:
-        sname: *sname
+        CESM_output_dir: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/
```
TeaganKing, line 90: We also talked about CESM_output_dir potentially being a global variable-- do we want to set that up now or in another PR?

examples/nblibrary/ocean_surface.ipynb
```diff
@@ -46,38 +46,41 @@
...
-    "DOUT_S = dcase.get_value('DOUT_S')\n",
-    "if DOUT_S:\n",
-    "  OUTDIR = dcase.get_value('DOUT_S_ROOT')+'/ocn/hist/'\n",
-    "else:\n",
-    "  OUTDIR = dcase.get_value('RUNDIR')+'/'\n",
-    "\n",
-    "print('Output directory is:', OUTDIR)\n",
-    "print('Casename is:', dcase.casename)"
+    "OUTDIR = f'{CESM_output_dir}/{Case}/ocn/hist/'\n",
+    "print('Output directory is:', OUTDIR)"
```
TeaganKing, line 83: Do we want to use log files rather than console printing here?

TeaganKing (approved):

TeaganKing:
I just made a few comments that could be addressed either in this PR or at a later time. Overall this looks good to me, though!

examples/nblibrary/ocean_surface.ipynb
mnlevy1981: I think `print()` is okay for inside notebooks, but if the other notebooks are using log files we should definitely be consistent. My vote is to leave it as-is, and to ensure consistency in a future PR

examples/coupled_model/config.yml
mnlevy1981: I'd like there to be an example of using `global_params`, so I'll make this change before merging the PR. And while I'm at it, I can clean up the unnecessary import statements :)

TeaganKing: That sounds great! Thanks Mike!

examples/nblibrary/ocean_surface.ipynb
TeaganKing: Sounds like a good solution to me!

README.md
```diff
@@ -66,10 +66,13 @@ After the last step is finished, you can use Jupyter to view generated notebooks
...
 directory to your local machine and look at `index.html` in a web browser.
 
-For users running on the NCAR super computers (derecho or casper), you can visualize the web page in a browser with the following steps:
-1. open a new browser window that points to https://fastx.ucar.edu:3300/session/
-1. open a default desktop icon
-1. select the browser client
-1. type `xterm` and hit enter to open a terminal
-1. In the terminal, run `cd ${CUPID_ROOT}/examples/coupled_model/computed_notebooks/quick-run/_build/html` to enter the `html` directory
-1. From the updated directory, run `firefox index.html &` to open a web browser pointed at the generated web page
+### Looking at Output
```
rmshkv, line 69: I mentioned this to you a while ago but an easier way is just to open the HTML files right in Jupyterhub - it might be a little more flaky after the update, but in the past I've had almost everything render properly.

examples/coupled_model/config.yml
```diff
@@ -57,27 +56,39 @@ compute_notebooks:
   ### notebook from nb_path_root, minus the .ipynb
   index:
     parameter_groups:
-      none:
-        sname: *sname
+      none: {}
```
rmshkv, line 59: I didn't realize you could just pass an empty dict, that's good to know! Quick note - the global params automatically get passed to all the notebooks, which looks a bit messy in a notebook that doesn't need them like index.ipynb. Probably worth addressing in the CUPiD code later on.

rmshkv (approved):

examples/coupled_model/config.yml
mnlevy1981: First I tried something like `none: null` to pass `None` instead of a dictionary, but got an error about how it needed to be a dictionary instead of a `NoneType` object.

I'll open an issue ticket and we can explore, but for now the parameter cell in `index.html` isn't too terrible.

README.md
mnlevy1981: Maybe I should mention this rendering issue on Zulip? Even clicking "Trust HTML" doesn't bring it up in a readable manner... I'm hesitant to mention looking at in directly on the Hub when the HTML rendering seems so finicky.

rmshkv: Ah that's unfortunate, I was hoping that rendering issue was a one-off thing. Definitely worth mentioning on Zulip since it used to work before the update.

PR #64, opened 2024-02-15
Add links to readme in documentation

TeaganKing:
Temporary fix for currently empty documentation

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,3 +6,5 @@ CUPiD Documentation
    :hidden:
 
     Installation Guide <install.rst>
+
+More details can be found in the main repository [README](https://github.com/TeaganKing/CUPiD/blob/main/README.md).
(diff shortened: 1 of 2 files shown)


TeaganKing:
@rmshkv this addresses the [currently empty installation instructions page](https://ncar.github.io/CUPiD/index.html) that you noticed!

rmshkv (approved):
This works! I also tried directly inputting the readme file in index.rst (`<../README.md>`) but that didn't seem to work at first - we may want to play around with that more later

TeaganKing:
> This works! I also tried directly inputting the readme file in index.rst (`<../README.md>`) but that didn't seem to work at first - we may want to play around with that more later

Sounds good! I'll merge this now for a short-term solution, but it would be good to have the README fully linked there as you suggest so that we can update one and not have to update the other...

TeaganKing:
(I am updating #38 to include Lev's suggestion)

PR #65, opened 2024-02-15
Embed README in index.rst

anissa111:
Related to: https://github.com/NCAR/CUPiD/issues/38#issuecomment-1946567725

Summary of changes:
- fixes some small warnings/typos in `README.md`
- embeds `README.md` in `docs/index.rst` by copying `CUPiD/README.md` into the docs directory in `docs/conf.py` and removes the logo in the top header because having an image in the heading puts it in the sidebar toc tree, which looks really weird.
- Adds copied `docs/README.md` to the `.gitignore` because it should only be created/edited automatically during documentation generation and not actually committed to the repo
- Removes separate installation page for now, since the information is on the home page

See local results:
image


rmshkv (approved):
Agreed with Teagan - this looks fantastic, thanks so much for setting it up @anissa111

TeaganKing:
Just to clarify, is the docs/README automatically generated and added to the documentation any time the normal README gets updated? If it's actually a copy that needs to be manually updated, maybe we could adjust that; if it's automatically maintained then I think this should be good to merge in...

docs/index.rst
```diff
@@ -5,6 +5,5 @@ CUPiD Documentation
    :maxdepth: 2
    :hidden:
 
-    Installation Guide <install.rst>
-
-More details can be found in the main repository `README <https://github.com/TeaganKing/CUPiD/blob/main/README.md>`_ .
+.. include:: README.md
```
TeaganKing, line 8: I think this means line shows that this is pulling in README.md as it is generated on the main page, rather than creating a duplicate. So, I think this is good to merge in. Does that make sense to you, @rmshkv ?

docs/conf.py
```diff
@@ -12,10 +12,27 @@
...
+# Copy README into docs
+# This is to allow us to remove the header image from the docs copy of README
+# without affecting the original README, but still pull the source README
+# into the docs build fresh each time.
+os.system('cp ../README.md ./README.md')
+
+# Remove any images from the first line of the README
+with open('README.md', 'r') as f:
+    readme1 = f.readline()
+    readme1 = re.sub('<.*?> ', '', readme1)
```
mnlevy1981, line 30: Does it make sense to be more specific in your regular expression? Would 

```python
readme1 = re.sub('<img.*?> ', '', readme1)
```

work for removing images from this line while allowing us to still include raw html at a later date? I don't work with markdown much, so maybe it doesn't matter much, but if we wanted to wrap all of the `CUPiD Logo CUPiD: CESM Unified Postprocessing and Diagnostics` in HTML (maybe we add a link to the website now that we have documentation?) then the current regular expression will strip out the whole line

mnlevy1981 (changes requested):

docs/index.rst
anissa111: Hi! Sorry I missed your earlier comment. Yes, it is pulling the README.md each time the documentation is generated. It does technically duplicate the file, but only for the duration of the documentation creation, it's not something that has to be maintained in parallel or anything.

TeaganKing: Fantastic! That sounds like a great strategy.

By the way, @mnlevy1981 mentioned that his last comment was a somewhat minor thing-- if it's easy to address this morning, great, if not, maybe we can just merge this in as is so that we can get [Lev's PR](https://github.com/NCAR/CUPiD/pull/62) in today as well.

docs/conf.py
anissa111: Sure! I currently can't think of a scenario where you'd want to use html instead of markdown syntax except for images, but there's no harm in being more specific!

mnlevy1981 (approved):

PR #67, opened 2024-02-21
Add link to README to sidebar

mnlevy1981:
Link text is "Home"

### DIFF
--- a/docs/home.rst
+++ b/docs/home.rst
@@ -1,0 +1,5 @@
+CUPiD Documentation
+===================
+
+.. include:: README.md
+   :parser: myst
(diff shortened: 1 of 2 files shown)


mnlevy1981:
If merged, the sidebar would look like this:

new sidebar

rmshkv (approved):
Looks good now!

anissa111 (changes requested):
Sorry I wasn't quick enough to catch this before, there's actually a way easier way to accomplish this.

In `config.py`, setting `home_page_in_navbar` is what you probably  want to do here:

https://github.com/NCAR/CUPiD/blob/main/docs/conf.py#L117

PR #78, opened 2024-03-06
Add single-variable time series file generation function from ADF

nusbaume:
This should hopefully generalize the timeseries generation capability so that any CUPiD script or notebook can use it.


cupid/timeseries.py
```diff
@@ -0,0 +1,325 @@
...
+
+        # Get a list of data variables in the 1st hist file:
+        hist_file_var_list = list(hist_file_ds.data_vars)
+        # Note: could use `open_mfdataset`, but that can become very slow;
+        #      This approach effectively assumes that all files contain the same variables.
+
+
+        # Check what kind of vertical coordinate (if any) is being used for this model run:
+        # ------------------------
+        if "lev" in hist_file_ds:
```
TeaganKing, line 145: Maybe adjust 'lev' if using cam vs clm, etc. Or pre-process files?

cupid/timeseries.py
```diff
@@ -0,0 +1,325 @@
+def create_time_series(case_names, hist_str, hist_locs, ts_dir, ts_done, overwrite_ts, start_years, end_years):
+    """
```
TeaganKing, line 2: If we want to process all of the variables, it would be nice to have a 'process-all' Boolean parameter to control that.

cupid/timeseries.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        return subprocess.run(cmd, shell=False)
+
+    # End def
+
+    # Notify user that script has started:
+    print("\n  Generating time series files...")
+
+    # get info about variable defaults
+    res = self.variable_defaults
```
TeaganKing, line 42: We will need to update `self`

TeaganKing: General cupid utility functions that would be useful may include `derive_variable`, etc

cupid/timeseries.py
wwieder: this may be abstract enough to get the clm coordinate dimensions of levsoi levgrnd, levdcmp, levurb, levlak, etc?  I wonder if a more generic approach may be to copy all the coordinate dimensions from a history file onto the single variable time series?

cupid/timeseries.py
```diff
@@ -0,0 +1,290 @@
+def create_time_series(case_names, hist_str, hist_locs, ts_dir, ts_done, overwrite_ts, start_years, end_years):
+    """
+    Generate time series versions of the history file data.
+
+    Args
+    ----
+     - case_names: list, str
+         name of simulaton case
+     - hist_str: str
+         CESM history number, ie h0, h1, etc.
```
wwieder, line 10: Is this generic enough for different history file types (e.g. patch or landunit level output from CLM, as opposed to grid cell averages)

cupid/timeseries.py
TeaganKing, line 42: Currently ADF uses a yaml file to define the variable defaults. Since variables will be different across components, we should probably have a similar default var yaml file for each component and create a CUPiD utility to load that similarly to the utility in [ADF's adf_obs](https://github.com/NCAR/ADF/blob/b96152255f7d386de34587eb8ed389f17db485b3/lib/adf_obs.py#L73).

cupid/timeseries.py
```diff
@@ -0,0 +1,290 @@
...
+        # Create path object for the history file(s) location:
+        starting_location = Path(hist_locs[case_idx])
+
+        # Check that path actually exists:
+        if not starting_location.is_dir():
+            emsg = "Provided 'cam_hist_loc' directory '{starting_location}' not found."
+            emsg += " Script is ending here."
+            # End if
+
+            self.end_diag_fail(emsg)
```
TeaganKing, line 69: We could implement a common utility similar to ADF's [end_diag_fail](https://github.com/NCAR/ADF/blob/b96152255f7d386de34587eb8ed389f17db485b3/lib/adf_base.py#L71). We should determine whether we want just a single component to exit or the entire CUPiD process to exit.

cupid/timeseries.py
```diff
@@ -0,0 +1,290 @@
...
+                + ["-o", ts_outfil_str]
+            )
+
+            # Add to command list for use in multi-processing pool:
+            list_of_commands.append(cmd)
+
+        # End variable loop
+
+        # Now run the "ncrcat" subprocesses in parallel:
+        with mp.Pool(processes=self.num_procs) as mpool:
```
TeaganKing, line 278: Maybe we can just pass num_procs to the timeseries function as a parameter

cupid/timeseries.py
```diff
@@ -0,0 +1,290 @@
...
+            _ = mpool.map(call_ncrcat, list_of_commands)
+
+        if vars_to_derive:
+            self.derive_variables(
```
TeaganKing, line 282: In ADF, this is defined [here](https://github.com/NCAR/ADF/blob/b96152255f7d386de34587eb8ed389f17db485b3/lib/adf_diag.py#L1011). This will be component-specific but we can pull out some common utility.

cupid/timeseries.py
TeaganKing: We can add a list of variables to the `config.yaml` file for each component under a `timeseries` key.

TeaganKing:
We should implement this by adding a call to `run.py` under `if time_series`.

cupid/timeseries.py
nusbaume: Just wanted to call attention to a draft PR @justin-richling opened in the ADF here:

https://github.com/NCAR/ADF/pull/278

It seems like we could potentially use the `sympy` package to allow users to write their own equations for derived variables without having to write new python code for each individual derived quantity.  Of course we haven't fully tested it yet so not sure if there are any dealbreakers, and it would bring in another few python packages to the conda environment.  Still, just wanted to give folks a heads up in case this seems like a feasible path forward for CUPiD.

TeaganKing: Thanks for mentioning this here! That seems like it could be really useful for allowing derived variables from each component!

cupid/timeseries.py
TeaganKing: We can do this via the `diag_var_list` argument, which will pull from the `config.yaml` file.

cupid/timeseries.py
TeaganKing: The details used for plotting seem to be ADF-specific, so after discussion with @mnlevy1981 , we're thinking that we can just supply a list of variables to the `config.yaml` file instead of adding new `.yaml` files with lots of extra details.

dabail10:
Is this ready for testing?

TeaganKing:
> Is this ready for testing?

Hi @dabail10 , sorry for the slow response, I'm in full-day training every day this week... I'm hoping to, at the least, wrap this up before our meeting next week, and will remove the draft label once it's ready for testing/review.

TeaganKing:
I've added the general infrastructure for components to use this timeseries generation. Each component will need to adjust the variables they want to generate (unless processing all vars) and the relevant history string (h0 is the current default) in `config.yml`. Additionally, in `run.py`, each component will need to update the height dimension they are using ('lev' is currently the default); this is in run.py instead of the config file since I don't anticipate it changing/needing to be modified by users.

cupid/timeseries.py
TeaganKing: @wwieder Are you referring to the hist_str here? This is now an editable quantity in the config.yml file. Would that be sufficient for those other kinds of history file output? I'm not familiar with those kinds of output, so any clarification would be helpful.

cupid/timeseries.py
TeaganKing: This has been moved to a new issue (#86) so that we can try to get this functionality merged in.

cupid/timeseries.py
TeaganKing: I mentioned this in a new issue (#86) so that we can keep the possibility of using `sympy` in mind going forward

cupid/run.py
```diff
@@ -7,37 +7,154 @@
...
-    # Abort if run with --time-series (until feature is added)
     if time_series:
         sys.tracebacklimit = 0
-        raise NotImplementedError("--time-series option not implemented yet")
+        with open("config.yml") as stream:
+            try:
+                config_contents = yaml.safe_load(stream)
+                config_timeseries_contents = config_contents["timeseries"]
+            except yaml.YAMLError as exc:
+                print(exc)
```
mnlevy1981, line 41: Should the `if time_series:` block get moved to after the

```py3
    # Get control structure
    control = cupid.util.get_control_dict(config_path)
    cupid.util.setup_book(config_path)
```

Then we'll only process the YAML file once. (Also, you can remove the `import sys` and `sys.tracebacklimit = 0` lines, I only included them to make the `NotImplementedError` exception easier to read)

cupid/run.py
```diff
@@ -7,37 +7,154 @@
...
+
+        # general timeseries arguments for all components
+        num_procs = config_timeseries_contents["num_procs"]
+
+        print("calling cam timeseries generation")
```
mnlevy1981, line 46: The rest of the print statements reference the general component, not the model itself (`lnd` instead of `clm`, `ocn` instead of `mom6`); I think you want `calling atm timeseries generation` here

cupid/run.py
```diff
@@ -7,37 +7,154 @@
...
+            config_timeseries_contents["ts_done"],
+            config_timeseries_contents["overwrite_ts"],
+            config_timeseries_contents["ocn_start_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.start_date
+            config_timeseries_contents["ocn_end_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.end_date
+            "lev", # TODO: ocean group will need to change this!
+            num_procs,
+            serial,
+        )
+
+        print("calling cice timeseries generation")
```
mnlevy1981, line 106: `ice` instead of `cice`?

examples/coupled_model/config.yml
```diff
@@ -48,6 +48,42 @@ global_params:
...
+  derive_vars_ocn: []
+  ocn_hist_str: 'h0'
+  ocn_start_years: [2]
+  ocn_end_years: [102]
+
+  cice_vars: ['process_all']
+  derive_vars_cice: []
+  cice_hist_str: 'h0'
+  cice_start_years: [2]
+  cice_end_years: [102]
```
mnlevy1981, line 79: `ice_*` instead of `cice_*` (and `derive_vars_ice` instead of `derive_vars_cice`)?

mnlevy1981 (changes requested):
This first set of comments is just from reading through the changes to `run.py` and `config.yml`; I'll take a closer look at `timeseries.py` and actually run the branch later today

examples/coupled_model/config.yml
TeaganKing: Making a new commit to address this...

cupid/run.py
TeaganKing: Done!

TeaganKing:
Thanks for those suggestions, @mnlevy1981 ! All of those comments have been addressed.

examples/coupled_model/config.yml
```diff
@@ -48,6 +48,42 @@ global_params:
   lc_kwargs:
     threads_per_worker: 1
 
+timeseries:
+  num_procs: 8
```
mnlevy1981, line 52: Ideally we can determine `num_procs` the same way `dask` does when spinning up a `LocalCluster`, but for now can the default be 12 to match the recommendation in the [NCAR tips](https://ncar.github.io/CUPiD/NCARtips.html)

cupid/timeseries.py
```diff
@@ -0,0 +1,416 @@
...
+
+        if vars_to_derive != []:
+            if component == "cam":
+                derive_cam_variables(
+                    vars_to_derive=vars_to_derive, ts_dir=ts_dir[case_idx]
+                )
+
+        if serial:
+            call_ncrcat(list_of_commands)
+        if not serial:  # if not serial
```
mnlevy1981: This could be an `else:`

mnlevy1981 (changes requested):
I'll run this branch before the 3:00 CUPiD meeting, but these are two more comments from just looking through the code

examples/coupled_model/config.yml
TeaganKing: See #86

cupid/run.py
```diff
@@ -2,42 +2,161 @@
...
+            [global_params["CESM_output_dir"]+'/'+timeseries_params['case_name']+'/glc/proc/tseries/'],
+            # Note that timeseries output will eventually go in /glade/derecho/scratch/${USER}/archive/${CASE}/${component}/proc/tseries/
+            timeseries_params["ts_done"],
+            timeseries_params["overwrite_ts"],
+            timeseries_params["glc_start_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.start_date
+            timeseries_params["glc_end_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.end_date
+            "lev", # TODO: glc group will need to change this!
+            num_procs,
+            serial,
+        )
```
mnlevy1981, line 150: Could this block be replaced with a loop?

```py3
for component in ['atm', 'ocn', 'lnd', 'ice', 'glc']:
        print(f"calling {component} timeseries generation")
        cupid.timeseries.create_time_series(
            component,
            timeseries_params[f"{component}_vars"],
            timeseries_params[f"derive_vars_{component}"],
            [timeseries_params["case_name"]],  # could also grab from compute_notebooks section of config file
            timeseries_params[f"{component}_hist_str"],
            [global_params["CESM_output_dir"] + "/" + timeseries_params["case_name"] + f"/{component}/hist/"],  # could also grab from compute_notebooks section of config file
            [global_params["CESM_output_dir"]+'/'+timeseries_params['case_name']+f'/{component}/proc/tseries/'],
            # Note that timeseries output will eventually go in /glade/derecho/scratch/${USER}/archive/${CASE}/${component}/proc/tseries/
            timeseries_params["ts_done"],
            timeseries_params["overwrite_ts"],
            timeseries_params[f"{component}_start_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.start_date
            timeseries_params[f"{component}_end_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.end_date
            "lev", # TODO: glc group will need to change this!
            num_procs,
            serial,
        )
```

It looks like we need to create a dictionary of vertical coordinates so we can do

```diff
- "lev", # TODO: glc group will need to change this!
+ vert_dim_name[component],
```

mnlevy1981 (changes requested):
Could we update `README.md` to mention the ability to generate time series files? Relatedly -- when I ran this myself, I didn't have the `nco` module loaded so I got an error

```
FileNotFoundError: [Errno 2] No such file or directory: 'ncrcat'
```

Maybe a quick note in the README that time series generation requires NCO? Or can we install NCO via conda? if so, maybe we should just add it to conda-dev?

kafitzgerald:
You can certainly install it via [conda](https://anaconda.org/conda-forge/nco), but I'm not sure if there are advantages to using the module(s) available on HPC.  Potentially some performance / compatibility tradeoffs there.

cupid/run.py
TeaganKing: I also updated a few places where this expects 'cam' instead of 'atm' for consistency (but left checks for eg cam in hist files)

TeaganKing:
Ok, I added another note to #86 to investigate any potential tradeoffs with NCO, as well as made a note in the README.

And thanks for the loop suggestion Mike-- definitely a good idea!

TeaganKing:
I also updated the project vision as I was looking at the README anyways...

rmshkv:
To elaborate on something I mentioned in https://github.com/NCAR/CUPiD/pull/88 - a potential suggestion (that we definitely don't have to take, definitely backseat driving a bit here): the CUPiD/ploomber framework is already set up to support multiple phases of tasks that depend on each other, e.g. running timeseries generation first before diagnostics. I would encourage at some point moving the timeseries config block to be under `compute_scripts` in `config.yml`, which is already set up to be organized by components that can be turned on or off, and parameters specific to each component that can be passed in to a universal script. This keeps things more generalizable and less hard-coded. We could then use options in ploomber's tasks to specify that timeseries should be run first. Before that happens we need to document the script functionality (which I'll do soon :) ) and maybe build it out a little more, but after that I'd be happy to make a draft of that modification and see if it's something we want to go with.

TeaganKing:
> To elaborate on something I mentioned in #88 - a potential suggestion (that we don't have to take, definitely backseat driving a bit here): the CUPiD/ploomber framework is already set up to support multiple phases of tasks that depend on each other, e.g. running timeseries generation first before diagnostics. I would encourage at some point moving the timeseries config block to be under `compute_scripts` in `config.yml`, which is already set up to be organized by components that can be turned on or off, and parameters specific to each component that can be passed in to a universal script. This keeps things more generalizable and less hard-coded. We could then use options in ploomber's tasks to specify that timeseries should be run first. Before that happens we need to document the script functionality (which I'll do soon :) ) and maybe build it out a little more, but after that I'd be happy to make a draft of that modification and see if it's something we want to go with.

@rmshkv thanks for this suggestion! Just to clarify, are you suggesting creating a new `compute_scripts` block? I don't see a pre-existing one. Also, maybe it would be best to do this after #88 comes in? I think it would be a fairly simple change to move these and adjust how things are called from the config file in run.py, but I'm a bit hesitant to do that _before_ #88 comes in.

cupid/timeseries.py
TeaganKing: Also mentioned in #86 ...

rmshkv:
> > To elaborate on something I mentioned in #88 - a potential suggestion (that we don't have to take, definitely backseat driving a bit here): the CUPiD/ploomber framework is already set up to support multiple phases of tasks that depend on each other, e.g. running timeseries generation first before diagnostics. I would encourage at some point moving the timeseries config block to be under `compute_scripts` in `config.yml`, which is already set up to be organized by components that can be turned on or off, and parameters specific to each component that can be passed in to a universal script. This keeps things more generalizable and less hard-coded. We could then use options in ploomber's tasks to specify that timeseries should be run first. Before that happens we need to document the script functionality (which I'll do soon :) ) and maybe build it out a little more, but after that I'd be happy to make a draft of that modification and see if it's something we want to go with.
> 
> @rmshkv thanks for this suggestion! Just to clarify, are you suggesting creating a new `compute_scripts` block? I don't see a pre-existing one (although I do see `compute_notebooks`). Also, maybe it would be best to do this after #88 comes in? I think it would be a fairly simple change to move these and adjust how things are called from the config file in run.py, but I'm a bit hesitant to do that _before_ #88 comes in.

The functionality to handle a `compute_scripts` block provided by `config.yml` already fully exists - there just isn't a `compute_scripts` block in the existing example in this repo. That's why I need to document it, it's kind of a secret feature right now :) And yeah agreed that this and #88 should be merged in first - I just wanted to get my thoughts written down somewhere so we can discuss.

TeaganKing:
@rmshkv Got it, that makes sense now! Thank you for clarifying! It seems like your suggestion is probably the best way forward, but I summarized this and some other relevant discussions that Mike and I had in #86 (with a note that we should probably use `compute_scripts`) just so the information is all in one place.

examples/coupled_model/config.yml
```diff
@@ -48,6 +48,47 @@ global_params:
...
+  ts_done: [False]
+  overwrite_ts: [False]
+  case_name: 'b.e23_alpha16b.BLT1850.ne30_t232.054'
+
+  atm_vars: ['ACTNI', 'ACTNL', 'ACTREI','ACTREL','AODDUST']
+  derive_vars_atm: [] # {'PRECT':['PRECL','PRECC'], 'RESTOM':['FLNT','FSNT']}
+  atm_hist_str: 'h0'
+  atm_start_years: [2]
+  atm_end_years: [102]
+  atm_level: 'lev'
```
mnlevy1981, line 62: Should the component-specific blocks use common variable names? E.g.

```yaml
atm:
  vars: ['ACTNI', 'ACTNL', 'ACTREI','ACTREL','AODDUST']
  derive_vars: [] # {'PRECT':['PRECL','PRECC'], 'RESTOM':['FLNT','FSNT']}
  hist_str: 'h0'
  start_years: [2]
  end_years: [102]
  level: 'lev'

lnd:
  vars: ['ALTMAX', 'COST_NACTIVE', 'DENIT', 'EFLX_DYNBAL']  #['process_all']
  derive_vars: []
  hist_str: 'h0'
  start_years: [2]
  end_years: [102]
  level: 'lev'
```

This would probably clean up the call to `cupid.timeseries.create_time_series()` because we could replace a bunch of individual variables with `**timeseries_params[component]`. I think it would also be more in line with what @rmshkv has been working on in #88

cupid/run.py
```diff
@@ -2,42 +2,82 @@
...
+            [global_params["CESM_output_dir"]+'/'+timeseries_params['case_name']+f'/{component}/proc/tseries/'],
+            # Note that timeseries output will eventually go in /glade/derecho/scratch/${USER}/archive/${CASE}/${component}/proc/tseries/
+            timeseries_params["ts_done"],
+            timeseries_params["overwrite_ts"],
+            timeseries_params[f"{component}_start_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.start_date
+            timeseries_params[f"{component}_end_years"],  # could get from yaml file in adf_quick_run.parameter_groups.none.config_fil_str, or for other notebooks config files, eg ocean_surface.parameter_gropus.none.mom6_tools_config.end_date
+            timeseries_params[f"{component}_level"],
+            num_procs,
+            serial,
+        )
```
mnlevy1981, line 71: Per my other comment from the review, I think it would be helpful to make the call here with named arguments

```py3
            cupid.timeseries.create_time_series(
            component=component,
            diag_var_list=timeseries_params[f"{component}_vars"],
            derive_vars=timeseries_params[f"derive_vars_{component}"],
            case_names=[timeseries_params["case_name"]],  # could also grab from compute_notebooks section of config file
```

Because then we could replace a bunch of the named arguments with `**time_series_params["component"]`

mnlevy1981 (changes requested):
This looks great! One more minor-ish suggestion, broken into two inline comments (and if we want to push it to #86 that's okay).

examples/coupled_model/config.yml
mnlevy1981: note that these keys would need to match the arguments expected in `create_time_series`, so `vars` would become `diag_var_list` (or the argument name could change in `timeseries.py`). I didn't look closely at each key in this dictionary, some others may need to change as well

cupid/run.py
TeaganKing: This suggestion was moved to #86

examples/coupled_model/config.yml
TeaganKing: The component-specific and standard var names are now implemented; thanks for the suggestion!

examples/coupled_model/config.yml
```diff
@@ -48,6 +48,52 @@ global_params:
...
+    derive_vars: []
+    hist_str: 'h0'
+    start_years: [2]
+    end_years: [102]
+    level: 'lev'
+
+  ocn:
+    vars: ['taux', 'tauy'] # ['process_all']
+    derive_vars: []
+    hist_str: 'h'
```
mnlevy1981, line 76: this should be `h.frc` to find the MOM6 history file with `taux` and `tauy`.

Also, one more thing for #86: MOM6 splits variables across a larger number of history files than other components, so at some point we may need to support `hist_str` being a list (which probably requires renaming it, since it's no longer required to be a `str`).

mnlevy1981 (approved):
Looks good, great work!

TeaganKing:
Thanks for all your help and suggestions from everyone who commented!

PR #79, opened 2024-03-06
Use click to allow command line arguments

### LINKED ISSUE #72: Add flags to cupid-run
It would be great to include flags in `cupid-run` for running in serial. Default will probably be to use parallelized setup, but there should be an option to request to run in serial.

mnlevy1981:
Sorry about the misleading branch name, but `click` seems better suited to this task than `argparse` since `pyproject.toml` wants to call `run()` from `cupid/run.py` rather than running `run.py` as a command line script.

Fixes #72


mnlevy1981:
```
$ cupid-run -h
Usage: cupid-run [OPTIONS] CONFIG_PATH

  Main engine to set up running all the notebooks.

Options:
  -s, --serial        Do not use LocalCluster objects
  -ts, --time-series  Run time series generation scripts prior to diagnostics
  -h, --help          Show this message and exit.
```

A few notes:

1. `cupid-run config.yml` will create a `LocalCluster` in the ocean and sea ice notebooks
2. `cupid-run --serial config.yml` will not create the `LocalCluster`
3. `cupid-run --time-series config.yml` is not supported:
```
   (cupid-dev) $ cupid-run --time-series config.yml
   NotImplementedError: --time-series option not implemented yet
   (cupid-dev) $
```

TeaganKing:
Should we also include the flag options in the README, or at least mention that descriptions are available with `-h`?

mnlevy1981:
> Should we also include the flag options in the README, or at least mention that descriptions are available with `-h`?

Yes! I'll update the README now.

rmshkv (approved):
This is a good first pass at these features, thanks for diving into this Mike!

Another thought is we probably want to include some more extensive documentation for people developing notebooks about including the serial parameter and logic with the local cluster, maybe in a separate page about parallelization, but I don't think that has to happen with this PR.

TeaganKing (approved):
Great! All these options are working for me. Thanks for adding a bit of info on this to the README. I agree with Lev's comment above, too.

Issue #83, opened 2024-03-13
Incompatibility with recent ADF update


rmshkv:
**Describe the bug**
Recent updates to `create_TEM_files.py` in ADF seem to be causing an incompatibility with how CUPiD is calling ADF. When versions of ADF after https://github.com/NCAR/ADF/commit/0d2b226bfddb9b7445685eed5091ada528e1c052 are used, the call to `adf.create_climo()` in cell 12 of `adf_quick_run.ipynb` raises the following error:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[12], line 2
      1 #Create model climatology (climo) files.
----> 2 adf.create_climo()

File ~/codes/CUPiD-ncarfork/examples/nblibrary/../../externals/ADF/lib/adf_diag.py:706, in AdfDiag.create_climo(self)
    703         self.end_diag_fail(emsg)
    705     # Run the listed scripts:
--> 706     self.__diag_scripts_caller(
    707         "averaging", avg_func_names, log_section="create_climo"
    708     )
    710 else:
    711     # If not, then notify user that climo file generation is skipped.
    712     print(
    713         "\n  No climatology files were requested by user, so averaging will be skipped."
    714     )

File ~/codes/CUPiD-ncarfork/examples/nblibrary/../../externals/ADF/lib/adf_diag.py:300, in AdfDiag.__diag_scripts_caller(self, scripts_dir, func_names, default_kwargs, log_section)
    297     self.debug_log(dmsg)
    299 # Call function
--> 300 self.__function_caller(
    301     func_name, func_kwargs=func_kwargs, module_name=func_name
    302 )

File ~/codes/CUPiD-ncarfork/examples/nblibrary/../../externals/ADF/lib/adf_diag.py:338, in AdfDiag.__function_caller(self, func_name, func_kwargs, module_name)
    335     return func(self, **func_kwargs)
    337 # Otherwise just run function as-is, and return result:
--> 338 return func(self)

File ~/codes/CUPiD-ncarfork/externals/ADF/lib/../scripts/averaging/create_TEM_files.py:210, in create_TEM_files(adf)
    206     adf.end_diag_fail(emsg)
    207 #End if
    208 
    209 #Get full path and file for file name
--> 210 output_loc_idx = tem_locs[case_idx]
    212 #Check if re-gridded directory exists, and if not, then create it:
    213 if not output_loc_idx.is_dir():

IndexError: list index out of range

```
**To Reproduce**
[...]

PR #84, opened 2024-03-13
Pin ADF to hash 4625c8a

mnlevy1981:
Later commits to ADF break the adf_quick_run notebook

This is a temporary workaround for #83 while we figure out what the permanent fix should be

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,6 +1,6 @@
 [ADF]
 local_path = externals/ADF
-branch = main
+hash = 4625c8a
 protocol = git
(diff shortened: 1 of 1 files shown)


mnlevy1981:
To test, do the following:

```bash
$ rm -rf externals/ADF/
$ ./manage_externals/checkout_externals
```

Then run the `coupled_model` example to verify that there are no errors in the atmosphere notebook

rmshkv (approved):
Tested it and it works for me! Thanks for the quick workaround!

PR #87, opened 2024-03-19
Update ADF to latest version

mnlevy1981:
There was a change to how the TEM computation was setup in the ADF configuration file; rather than a top-level tem_info key, there is now a block of tem variables under each case.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,6 +1,6 @@
 [ADF]
 local_path = externals/ADF
-hash = 4625c8a
+branch = main
 protocol = git
 repo_url = https://github.com/NCAR/ADF.git
 required = True
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,6 +1,6 @@
 [ADF]
 local_path = externals/ADF
-branch = main
+hash = b961522
 protocol = git
 repo_url = https://github.com/NCAR/ADF.git
 required = True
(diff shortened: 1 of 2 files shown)


mnlevy1981:
I thought I had this figured out, but when running the notebook with the latest ADF and the CUPiD code on this branch I got the following warning:

TEM warning

Running with the older version of ADF / CUPiD does not throw these warnings:

No TEM warning

mnlevy1981:
@nusbaume and @justin-richling -- I guess my big question is "are those `invalid value` warnings a sign that I've done something wrong on this branch?" If those are now expected warnings, or indicative of an ADF bug, then I can open this up for review / merging. The three plots generated by this notebook look fine, but I haven't tried a run where I regenerate the TEM files (I'll start that right now)

mnlevy1981:
I think that previous warning came because I was actually regenerating the TEM file; 4c9fedc updates `cam_tem_loc` to use the old file and the warning isn't present when reading it

Externals.cfg
```diff
@@ -1,6 +1,6 @@
 [ADF]
 local_path = externals/ADF
-hash = 4625c8a
+branch = main
```
nusbaume, line 3: I think right now `main` is updated fairly frequently.  Would it make sense to just grab the latest hash instead?  Alternatively we could make a new tag or branch of the ADF that is specifically for CUPiD.  Just let me know what you think would be best on your end.

nusbaume (commented):
The changes look good to me (and I agree those output differences were due to the ADF re-calculating the TEM files)!  However, I did have a possible concern about using ADF's `main` branch.  Maybe CUPiD would benefit from a more stable alternative?

Externals.cfg
mnlevy1981: I think eventually pointing to ADF tags makes the most sense, but maybe for now we should point to specific hashes? It does seem safer to lag behind the latest / greatest ADF knowing that we are using a compatible version, and I'd rather intentionally test a newer ADF hash when there are updates we want to incorporate than be surprised when a new CUPiD user pulls down the latest ADF `main` and it turns out to be incompatible...

mnlevy1981: @nusbaume I just pushed 59a620f which should get us back to pinning to a hash (but the most recent commit to `main` rather than the February commit)

nusbaume (approved):
Everything looks good to me now, thanks!

Issue #63, opened 2024-02-14
Change the cupid-run API?


mnlevy1981:
It would be nice to break up `examples/coupled_model/config.yml` into many smaller files, and then have a mechanism to pass multiple files to `cupid-run`. I see two big advantages to this:

1. It makes it much easier to run a single notebook instead of every notebook in `nblibrary/`
2. We could drop the `coupled_model/` directory altogether, and just have `examples/` contain a bunch of YAML files

I'm not sure what we would want to do about keys in `config.yaml` that apply to all notebooks (`data_sources`, `computation_config`, `global_params`) or the Jupyter Book generation. Maybe we could keep `config.yaml` for global settings, and then come up with a naming convention for the per-component files. If they are all `cupid-[component].yaml`, then something like

```
$ cupid-run --global-config config.yaml cupid-*.yaml
```

would be the equivalent of the current `cupid-run config.yaml` while allowing for `$ cupid-run --global-config config.yaml cupid-atm.yaml` if you only want to run the atmosphere notebook. (And maybe `config.yaml` is the default value for `--global-config` so it can be omitted most of the time?)

I like the general layout of this new API, though every option / file name mentioned was "first thing that came to mind" and could likely be improved upon with a little thought.

mnlevy1981:
Another suggestion was to add another layer to the `compute_notebooks` dictionary where we break the notebooks into groups by components:

```yaml
compute_notebooks:

  # This is where all the notebooks you want run and their
  ### parameters are specified. Several examples of different
  ### types of notebooks are provided.

  # The first key (here simple_no_params_nb) is the name of the
  ### notebook from nb_path_root, minus the .ipynb
  index:
    parameter_groups:
      none: {}

  atmosphere:
    adf_quick_run:
      parameter_groups:
        none:
          adf_path: ../../externals/ADF
          config_path: .
          config_fil_str: "config_f.cam6_3_119.FLTHIST_ne30.r328_gamma0.33_soae.001.yaml"

  ocean:
    ocean_surface:
      parameter_groups:
        none:
          Case: b.e23_alpha16b.BLT1850.ne30_t232.054
          savefigs: False
          mom6_tools_config:
            start_date: '0091-01-01'
            end_date: '0101-01-01'
            Fnames:
              native: 'mom6.h.native.????-??.nc'
              static: 'mom6.h.static.nc'
            oce_cat: /glade/u/home/gmarques/libs/oce-catalogs/reference-datasets.yml

  land:
[...]

Issue #36, opened 2023-12-20
nbscuid - Move file creation related to jupyter book out of run.py and into build.py


rmshkv:
https://github.com/NCAR/CUPiD/blob/main/nbscuid/util.py#L114C1-L168

Currently lots of things happen together in `setup_book()`, including copying over parameters related to jupyter book creation from config.yml to separate files. This creates confusion if you later run nbscuid-build config.yml having changed some parameters, because the files were already created when you ran nbscuid-run and will not carry over.

PR #88, opened 2024-03-19
API updates to specify components

rmshkv:
Addresses https://github.com/NCAR/CUPiD/issues/63.

This adds a layer of keys in `config.yml` that groups notebooks (and scripts) by ESM component, e.g. `atmosphere`, `ocean`, etc. These keys are referenced by flags that can be passed in to `cupid-run`, e.g. `-atm`, `-ocn`, etc., as specified in the README. If no component flags are passed in, all components are run. This can also be done explicitly by using `--all` or `-a`.

This update also prompted changes to how we're managing the environment (kernel) checking step that was being done by `util.get_control_dict()` called by `cupid-run`. We still check the existence of all environments specified in `config.yml` (regardless of whether that component is turned on by flags), but just raise a warning if the environment does not exist. If that notebook is specified to be run, another warning is raised and that notebook is not run, but the others still are. Also, if neither a `default_kernel_name` nor a notebook-specific `kernel_name` is provided, `cupid-analysis` is assumed and another warning is raised.


rmshkv:
Also, I don't love how the warnings show up currently, you end up with a large block of text mixed in with all the Ploomber stuff and also the  `warnings.warn()`... code used to generate the warnings gets printed out too. Please feel free to propose a better way if you know one!

TeaganKing:
Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.

cupid/run.py
```diff
@@ -80,19 +91,50 @@ def run(config_path, serial=False, time_series=False):
...
     all_nbs = dict()
-
-    for nb, info in control['compute_notebooks'].items():
-
+    for nb, info in control['compute_notebooks']['infrastructure'].items():
         all_nbs[nb] = info
-
+        
+    # automatically run all if not specified
+    if True not in [atmosphere, ocean, land, seaice, landice]:
```
TeaganKing, line 107: Seems like a good default option!

cupid/run.py
```diff
@@ -105,11 +147,29 @@ def run(config_path, serial=False, time_series=False):
...
+                            all_scripts[script] = info
+                    else:
+                        warnings.warn(f"No scripts for {comp_name} component specified in config file.")
+
+        # Checking for existence of environments
+    
+        for script, info in all_scripts.copy().items():
+            if not control["env_check"][info["kernel_name"]]:
+                bad_env = info["kernel_name"]
+                warnings.warn(f"Environment {bad_env} specified for {script}.py could not be found; {script}.py will not be run.")
```
TeaganKing, line 170: Super minor, but would it be useful to suggest checking the README for installation instructions here? It may be obvious, but could potentially help some users.

examples/coupled_model/config.yml
```diff
@@ -56,53 +56,55 @@ compute_notebooks:
...
+        parameter_groups:
+          none:
+            cases:
+              - ctsm51d159_f45_GSWP3_bgccrop_1850pAD
+              - ctsm51d159_f45_GSWP3_bgccrop_1850pSASU
+            type:
+              - 1850pAD
+              - 1850pSASU
+    seaice:
+      seaice:
```
TeaganKing, line 97: This structure is a lot clearer with your changes!

We may want to encourage names for notebooks that are slightly different from the component name (eg, `seaice` listed twice here is confusing). Perhaps this is beyond the scope of this PR, but I wonder if down the line it may be useful to have some standard notebook names (eg, obs comparisons, basic diagnostics, etc) across components...

rmshkv:
> Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.

Ah thanks for the heads up, I didn't realize how far along the timeseries generation work was! I think there are several different ways we could integrate these features...potentially even something like including the timeseries generation code for each component under `compute_scripts`, which already get turned on or off by the component flags but could then also be affected by the `-ts` flag? Maybe we can chat about it at the meeting tomorrow.

TeaganKing (commented):
I generally approve these updates; but yes, let's talk more about the `ts` integration tomorrow too!

cupid/util.py
```diff
@@ -77,37 +32,36 @@ def get_control_dict(config_path):
...
+    
+    for nb_category in control["compute_notebooks"].values():
+        for nb, info in nb_category.items():
+            if "kernel_name" not in info:
+                if default_kernel_name is not None:
+                    info["kernel_name"] = default_kernel_name
+                else:
+                    info["kernel_name"] = "cupid-analysis"
+                    warnings.warn(f"No environment specified for {nb}.ipynb; assuming cupid-analysis environment.")
+            env_check[info["kernel_name"]] = None
```
mnlevy1981, line 47: Can we do the

```py3
env_check[info["kernel_name"]] = info["kernel_name"] in jupyter_client.kernelspec.find_kernel_specs()
```

right away instead of looping through the `env_check` keys later? Maybe something like

```py3
if info["kernel_name"] not in env_check:
  env_check[info["kernel_name"]] = info["kernel_name"] in jupyter_client.kernelspec.find_kernel_specs()
```

to avoid checking for the same environment repeatedly?

cupid/util.py
```diff
@@ -77,37 +32,36 @@ def get_control_dict(config_path):
...
-            manage_conda_kernel(d["kernel_name"]).ensure_installed()
-        
+        for script_category in control["compute_scripts"].values():
+            for script, info in script_category.items():
+                if "kernel_name" not in info:
+                    if default_kernel_name is not None:
+                        info["kernel_name"] = default_kernel_name
+                    else:
+                        info["kernel_name"] = "cupid-analysis"
+                        warnings.warn(f"No environment specified for {script}.py; assuming cupid-analysis environment.")
```
mnlevy1981, line 57: Same comment as in the `control["compute_notebooks"]` loop above

cupid/util.py
```diff
@@ -77,37 +32,36 @@ def get_control_dict(config_path):
...
+            env_check[info["kernel_name"]] = None
+
+    for environment in env_check.keys():
+        env_check[environment] = environment in jupyter_client.kernelspec.find_kernel_specs()
```
mnlevy1981, line 61: Same comment as in the `control["compute_notebooks"]` loop above (setting `env_check` in 47 and 58 means we don't need the loop at 60 and 61)

cupid/util.py
```diff
@@ -77,37 +32,36 @@ def get_control_dict(config_path):
...
+
+    control["env_check"] = env_check
```
mnlevy1981, line 63: With the comments above, we can likely get rid of the local `env_check` and just populate `control["env_check"][info["kernel_name"]]`:

```py3
if info["kernel_name"] not in control["env_check"]:
  control["env_check"][info["kernel_name"]] = info["kernel_name"] in jupyter_client.kernelspec.find_kernel_specs()
```

cupid/util.py
```diff
@@ -77,37 +32,36 @@ def get_control_dict(config_path):
...
-        
-        for script, d in control["compute_scripts"].items():
-            assert "kernel_name" in d, f"kernel information missing for {script}.py"
-
-    for nb, d in control["compute_notebooks"].items():
-        manage_conda_kernel(d["kernel_name"]).ensure_installed()
-
+    env_check = dict()    
+    
+    for nb_category in control["compute_notebooks"].values():
```
mnlevy1981, line 39: For the `compute_scripts` block we have a

```py3
if "compute_scripts" in control:
```

do we need a similar

```py3
if "compute_notebooks" in control:
```

check here?

mnlevy1981 (changes requested):
I've noted a handful of small changes that would help clean things up a bit, and I also have a bigger request: could we create subdirectories in `examples/nblibrary` to match the new `compute_notebook` keys and move the appropriate notebooks in? This would potentially lead to some minor tweaks in the file name itself... So

```
index.ipynb -> infrastructure/index.ipynb
adf_quick_run.ipynb -> atmosphere/quick_run.ipynb
ocean_surface.ipynb -> surface_plots.ipynb
land_comparison.ipynb -> comparison_plots.ipynb
seaice.ipynb -> seaice/[something descriptive].ipynb
```

We would need JupyterBook to respect this directory structure; maybe `infrastructure/index.ipynb -> index.html` is a special case, but we'd want `ocean/surface_plots.html` in case other components also want to have a `surface_plots.ipynb` notebook.

cupid/util.py
rmshkv: Originally I was assuming that every collection would have notebooks but that scripts were optional, hence this logic, but that isn't necessarily how people are going to use this. I'll add the extra logic where necessary.

cupid/util.py
```diff
@@ -153,51 +101,48 @@ def setup_book(config_path):
 
     # get list of computational notebooks
     
-    nb_path_root = os.path.expanduser(control['data_sources']['nb_path_root'])
-    
-    compute_notebooks = [f"{nb_path_root}/{f}.ipynb" for f in control["compute_notebooks"].keys()]
+    # if 'compute_notebooks' in control:
 
-    # get toc files; ignore glob expressions
```
rmshkv, line 160: I commented out this whole block and think it would be fine to remove it - it seems like it's mostly additional checking that's handled fine by Jupyter Book itself, but I want some second opinions before I actually do that

examples/coupled_model/config.yml
```diff
@@ -56,53 +55,55 @@ compute_notebooks:
...
+              - g.e23_a16g.GJRAv4.TL319_t232_hycom1_N75.2024.005
+              - g.e23_a16g.GJRAv4.TL319_t232_zstar_N65.2024.004
+            begyr1: 245
+            endyr1: 305
+            begyr2: 245
+            endyr2: 305
+            nyears: 25
 
 
 ########### JUPYTER BOOK CONFIG ###########
```
rmshkv, line 109: For the Jupyter book section, I went with explicitly specifying the path with the new folder for each one, rather than building in official handling for the "chapter caption" to match the directory structure. I think this keeps a bit more flexibility and makes it more explicit what the Jupyter book is doing (+ avoids an additional layer of hard-codey custom stuff), but we could choose to go the other way if we want.

cupid/util.py
rmshkv: This also connects to more generally cleaning up setup_book as per https://github.com/NCAR/CUPiD/issues/36, but I'm going to save the rest of that for a different PR

examples/coupled_model/config.yml
TeaganKing: Putting this into an issue #96

TeaganKing (approved):
Thanks for all your work on this, @rmshkv ! I ran through the notebooks and generated a jupyter book, and this all runs smoothly as far as I can tell.

I think the explicit path specification in the Jupyter Book Config seems clear, and that seems fine to me to remove that commented block that you mentioned.

cupid/run.py
```diff
@@ -113,35 +120,59 @@ def run(config_path, serial=False, time_series=False):
...
+        
+        for nb, info in control['compute_notebooks']['infrastructure'].items():
+            all_nbs[nb] = info
+            all_nbs[nb]['nb_path_root'] = nb_path_root + '/infrastructure'
+            all_nbs[nb]['output_dir'] = output_dir + '/infrastructure'
+            
+        # Automatically run all if not specified
+        
+        if True not in [atmosphere, ocean, land, seaice, landice]:
+            all = True
```
mnlevy1981, line 143: Do we need a `--all` option if we are setting `all` here? What do want to have happen if a user runs `cupid-run --all --atmosphere config.yml`? If the answer is "a user shouldn't run with those options together", it makes more sense to drop `all` from `click` / the input argument list and just set

```py3
all = (True not in [atmosphere, ocean, land, seaice, landice])
```

mnlevy1981 (changes requested):
I didn't make it very far through the code, and looking at my calendar I might not have time to come back to this until Tuesday... but I'll try to add more comments sooner than that

cupid/run.py
```diff
@@ -113,35 +120,59 @@ def run(config_path, serial=False, time_series=False):
...
+        
+        if all:
+            for comp_name, comp_nbs in control["compute_notebooks"].items():
+                for nb, info in comp_nbs.items():
+                    all_nbs[nb] = info
+                    all_nbs[nb]['nb_path_root'] = nb_path_root + '/' + comp_name
+                    all_nbs[nb]['output_dir'] = output_dir + '/' + comp_name
```
mnlevy1981, line 150: Does this need to be in an `if all` block? Could we do something like

```py3
for comp_name, comp_bool in component_options.items():
  if comp_name in control['compute_notebooks']) and (comp_bool or all):
    for nb, info in control['compute_notebooks'][comp_name].items():
      all_nbs[nb] = info
      all_nbs[nb]['nb_path_root'] = nb_path_root + '/' + comp_name
      all_nbs[nb]['output_dir'] = output_dir + '/' + comp_name
  elif comp_bool:
        warnings.warn(f"No notebooks for {comp_name} component specified in config file.")
```

to encompass both the `if all` and `else` portions of the existing code? I think this snippet behaves exactly like what you've written, but there's only one loop over `control['compute_notebooks'][comp_name]` to update if we change the `config.yml` schema again.

I haven't tested the above block, so you might need a small tweak but the logic in the `if` / `else` block should be sound.

cupid/run.py
```diff
@@ -150,29 +181,40 @@ def run(config_path, serial=False, time_series=False):
...
         all_scripts = dict()
 
-        for script, info in control["compute_scripts"].items():
-
-            all_scripts[script] = info
-
+        # Automatically run all if not specified
+        
+        if True not in [atmosphere, ocean, land, seaice, landice]:
+            all = True
```
mnlevy1981, line 187: It looks we have an identical block of code in 142 / 143, but that's in an `if 'compute_notebooks' in control:` block and this is in `if "compute_scripts" in control:`. Could we move the first instance of this out of the if statement (maybe directly after defining `component_options`) and then remove this occurence?

cupid/run.py
```diff
@@ -150,29 +181,40 @@ def run(config_path, serial=False, time_series=False):
...
+            
+        if all:
```
mnlevy1981, line 189: Same comment as the `compute_notebooks` block - let's reformulate the if checks and only have a single loop through `control['compute_scripts'][comp_name]`

cupid/util.py
mnlevy1981: It sure looks like the commented out code is un-necessary -- I vote for removing it altogether (if we realize we need it later, we can grab it from an old commit)

mnlevy1981:
Most of my comments on this pass-through are related to reducing the amount of duplicated code. There is still a lot of similarities between `compute_notebooks` and `compute_scripts` sections, but cleaning that up is probably best left for a separate issue

cupid/run.py
mnlevy1981: Now I'm second-guessing the logic. I think we really want

```py3
if comp_name in control['compute_notebooks']) and comp_bool:
  ...
elif comp_bool and not all:
  ...
```
So if `comp_bool` is true and the component has notebooks defined in control structure, add those to `all_nbs`. If `comp_bool and not all` then the user specified `--comp` but we're in the else block so the component name does not have notebooks defined in the control structure and that's when we want to warn about requesting a specific component that isn't in the config file.

rmshkv:
> Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.

I just implemented this (-ts and component flags will only run those components). There's a tiny bit of extra code because the timeseries block references components as "atm", "ocn", etc and the notebooks/scripts blocks reference them as "atmosphere", "ocean", etc but it works fine and we can address it later, in the interest of getting this PR in.

rmshkv:
> > Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.
> 
> I just implemented this (-ts and component flags will only run those components). There's a tiny bit of extra code because the timeseries block references components as "atm", "ocn", etc and the notebooks/scripts blocks reference them as "atmosphere", "ocean", etc but it works fine and we can address it later, in the interest of getting this PR in.

On second thought, maybe we do want them to match...I'll make everything use the short names, and we can change that later if we decide to.

mnlevy1981:
> > > Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.
> > 
> > 
> > I just implemented this (-ts and component flags will only run those components). There's a tiny bit of extra code because the timeseries block references components as "atm", "ocn", etc and the notebooks/scripts blocks reference them as "atmosphere", "ocean", etc but it works fine and we can address it later, in the interest of getting this PR in.
> 
> On second thought, maybe we do want them to match...I'll make everything use the short names, and we can change that later if we decide to.

If you haven't made the change yet, I'd prefer changing the time series to use the longer names... but if you are already using the short names we can clean that up in a future PR :)

rmshkv:
> > > > Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.
> > > 
> > > 
> > > I just implemented this (-ts and component flags will only run those components). There's a tiny bit of extra code because the timeseries block references components as "atm", "ocn", etc and the notebooks/scripts blocks reference them as "atmosphere", "ocean", etc but it works fine and we can address it later, in the interest of getting this PR in.
> > 
> > 
> > On second thought, maybe we do want them to match...I'll make everything use the short names, and we can change that later if we decide to.
> 
> If you haven't made the change yet, I'd prefer changing the time series to use the longer names... but if you are already using the short names we can clean that up in a future PR :)

I made the change just a minute ago, but it's not too hard to change back later. I also prefer the long names for clarity, but I don't know much about what the timeseries code is doing and it looks like some of the directories it's creating might depend on the component shortname, so I didn't want to touch that myself.

rmshkv:
All right, I think that addresses everything requested! Feel free to run some tests again and let me know what you think.

TeaganKing:
> > > > > Hi @rmshkv , I'll look at this PR in more detail, but wanted to post a quick note as I am realizing that there will probably be some conflicts between this PR and #78. I think that's fine, and I can update #78 if this comes in first. We may also want to discuss whether the component flags also result in timeseries being run for only the specified components (eg, if you run `cupid-run config.yml -ts -atm`, that would generate timeseries and run the notebooks for `atm` only). I think this may be ideal and still seems clear to me, but I'm open to other thoughts there.
> > > > 
> > > > 
> > > > I just implemented this (-ts and component flags will only run those components). There's a tiny bit of extra code because the timeseries block references components as "atm", "ocn", etc and the notebooks/scripts blocks reference them as "atmosphere", "ocean", etc but it works fine and we can address it later, in the interest of getting this PR in.
> > > 
> > > 
> > > On second thought, maybe we do want them to match...I'll make everything use the short names, and we can change that later if we decide to.
> > 
> > 
> > If you haven't made the change yet, I'd prefer changing the time series to use the longer names... but if you are already using the short names we can clean that up in a future PR :)
> 
> I made the change just a minute ago, but it's not too hard to change back later. I also prefer the long names for clarity, but I don't know much about what the timeseries code is doing and it looks like some of the directories it's creating might depend on the component shortname, so I didn't want to touch that myself.

Just as an FYI, the time series code isn't super dependent on how the components are named; There's a comment in line 285 that would need to be updated, as well as just the names in lines 329, 52, and 232. I'm fine with keeping the short names for now, though.

TeaganKing:
All the tests I've run are looking good! I do also agree that updated the logic regarding 'all' is improved when the assumption is that 'all' is the default unless one component is specified, and a flag is not needed to specify 'all'. I think that the lack of a flag clarifies that 'all' is default instead of potential confusion with a default that can also be specified with a flag.

mnlevy1981:
The code in `run.py` looks great and `cupid-run config.yml` generates output as expected, but `cupid-build` isn't getting the sidebar right anymore:

cupid book

I was hoping the sidebar would have the name of all the components, and then links to each of the component's notebooks underneath.

I really like how this turned out:

```
(cupid-dev) examples/coupled_model$ cupid-run -atm -glc config.yml
/glade/work/mlevy/codes/CUPiD/cupid/run.py:152: UserWarning: No notebooks for glc component specified in config file.
```

(And `adf_quick_run.ipynb` was still executed).

rmshkv:
> The code in `run.py` looks great and `cupid-run config.yml` generates output as expected, but `cupid-build` isn't getting the sidebar right anymore:
> 
> cupid book
> 
> I was hoping the sidebar would have the name of all the components, and then links to each of the component's notebooks underneath.
> 
> I really like how this turned out:
> 
> ```
> (cupid-dev) examples/coupled_model$ cupid-run -atm -glc config.yml
> cupid/run.py:152: UserWarning: No notebooks for glc component specified in config file.
> ```
> 
> (And `adf_quick_run.ipynb` was still executed).

My bad, I just forgot to change the explicit paths in the jupyter book config with the new short folder names. Should be fixed now!

mnlevy1981:
> My bad, I just forgot to change the explicit paths in the jupyter book config with the new short folder names. Should be fixed now!

After updating to f114b83 I needed to rerun `cupid-run` and `cupid-build`, but the page looks great now!

cupid book 2

mnlevy1981 (approved):

Issue #85, opened 2024-03-15
cupid-clear command?


rmshkv:
It might be nice to have a simple command along the lines of `cupid-clear config.yml` that manually clears the computed_notebooks folder specified by a given config file. Ploomber's caching functionality is handy but doesn't seem to work the way we need for changes to executed notebooks, and I've found myself having to manually delete the output notebooks a lot to get them to rerun, so this might speed things up!

shivaniikum:
I am interested in working on this issue.

PR #91, opened 2024-04-07
"cupid-clear" Command Implementation Addressing Issue #85

shivaniikum:
**This pull request addresses issue #85 and contains the following additions:** 
1) A new clear.py script located at "cupid/clear.py" 
2) cupid-clear = "cupid.clear:clear" added to the last line of the pyproject.toml file

**The clear.py script can be run with the following steps:**

1) After adding these changes to your local environment, type "pip install -e ." in your terminal

2) Navigate to the directory where the config.yml file is located (this config.file contains the run_dir variable which is the path to the computed_notebooks folder you want to clear). For example: "cd examples/coupled_model"

3) Run the command "cupid-clear config.yml" and this should clear the computed_notebooks folder at the path specified by the run_dir variable in your config.yml file (if you're running cupid-clear from the examples/coupled_model directory and the config.yml located here has run_dir= . then the folder at the path examples/coupled_model/computed_notebooks will be cleared, if it exists)


cupid/clear.py
```diff
@@ -0,0 +1,52 @@
...
+            # If item is a directory, recursively clear it
+            elif os.path.isdir(itemPath):
+                clearFolder(itemPath)   
+        # After deleting all items, remove the folder itself
+        os.rmdir(folderPath)    
+        print(f"All contents in {folderPath} have been cleared.")
+    except Exception as e:
+        print(f"Error occurred while clearing contents of the file at path {folderPath}: {e}")
+        
+def readConfigFile(configFilePath):
```
TeaganKing, line 23: Within this function, it may be useful to keep a similar structure for reading the config file to that in `run.py`. So, instead of the `yaml.safe_load(file)`, you may want to use the pre-exisiting get_control_dict function as is used [here](https://github.com/shivaniikum/CUPiD/blob/7b2c7c4e31b1b2ff36d80e0510b739c60d17b1cb/cupid/run.py#L34) to read the contents of the configuration file, and then read in `control['data_sources']['run_dir']`.

TeaganKing (changes requested):
Thank you for contributing this feature! It looks like this is deleting all files under and including `computed_notebooks` when `cupid-clear config.yaml` is run, so it's working as expected, which is great!

When I set up the environment again from scratch, I was running into a few issues with the cupid-dev environment (not installing all packages properly with `mamba env create -f environments/dev-environment.yml`) that I'd like to double check once our machines are back up (hopefully soon).

One note, it may be useful to document this capability (just one line or so) in the README.md under the 'Running' section (before 'Cupid Options').

rmshkv (commented):
Agreed with Teagan's suggestions about the get_control_dict function and adding to the readme! I also want to test run this myself as soon as our machines are back up.

TeaganKing:
Also, just to be clear, I don't think any of my aforementioned environment issues are related to this PR since there were no changes to the environment yaml files-- I just wanted to be sure to note that and check to make sure we don't have other issues going on :)

shivaniikum:
Thank you for your feedback. I've modified clear.py to use the pre-exisiting get_control_dict function, then to read in control['data_sources']['run_dir']. I also added instructions about clear.py to the README.md. Please let me know if this works correctly.

TeaganKing:
Great! Thanks @shivaniikum . I'll test that again once our machines are back up!

TeaganKing (approved):
Ok, this is looking great! Thanks for making those changes.

The environment issues were specific to my machine and how I had set things up-- I've also tested on derecho (an NCAR machine) and everything looks good with the environment setup now.

The new method of finding the configuration file `run_dir` seems to be working smoothly.

I'll approve this PR, but I think @rmshkv wanted to test it out as well before we merge it in.

shivaniikum:
> Ok, this is looking great! Thanks for making those changes.
> 
> The environment issues were specific to my machine and how I had set things up-- I've also tested on derecho (an NCAR machine) and everything looks good with the environment setup now.
> 
> The new method of finding the configuration file `run_dir` seems to be working smoothly.
> 
> I'll approve this PR, but I think @rmshkv wanted to test it out as well before we merge it in.

That sounds awesome! Thank you!

cupid/clear.py
```diff
@@ -0,0 +1,53 @@
...
+#Entry point to this script
+def clear():
+    #Get the current working directory and add 'config.yml' to the path to 
+    #obtain the path to the config.yml file in the current working directory
+    currWorkingDir =  os.getcwd() 
+    configFilePath = os.path.join(currWorkingDir, 'config.yml')
+
+    run_dir = readConfigFile(configFilePath)
+
+    if run_dir:
```
mnlevy1981, line 52: Do we need this `if` statement? If `run_dir` is empty, `readConfigFile()` will raise a `ValueError`

cupid/clear.py
```diff
@@ -0,0 +1,53 @@
...
+    except Exception as e:
+        print(f"Error occurred while clearing contents of the file at path {folderPath}: {e}")
+
+def readConfigFile(configFilePath):
+    #Given the file path to config.yml, this function reads the config file content and 
+    #returns the val of the run_dir string with '/computed_notebooks' appended to it 
+    try:
+        #Obtain the contents of the config.yml file and extract the run_dir variable
+        control = cupid.util.get_control_dict(configFilePath)
+        run_dir = control['data_sources']['run_dir']
```
mnlevy1981, line 29: Since you are going to check to see if `run_dir` exists, I would use the `dict.get()` function and set `run_dir` to `None` or an empty string if it is not in `control['data_sources']`:

```py3
run_dir = control['data_sources'].get('run_dir', None)
```

mnlevy1981 (changes requested):
Just a few small changes, this looks great!

shivaniikum:
Thank you for reviewing my PR  @mnlevy1981. I've modified the clear.py script and README.md as you suggested (and also removed redundant exceptions for the ones I noticed python reports). Please let me know how this looks.

shivaniikum:
Thank you for reviewing my PR @mnlevy1981. I've modified the clear.py script and README.md as you suggested (and also removed redundant exceptions for the ones I noticed python reports). Please let me know how this looks.

mnlevy1981:
> Thank you for reviewing my PR @mnlevy1981. I've modified the clear.py script and README.md as you suggested (and also removed redundant exceptions for the ones I noticed python reports). Please let me know how this looks.

Thanks for making those changes! Could you add the `try` / `except` block that you removed from `readConfigFile()` to `get_control_dict()`?

shivaniikum:
> > Thank you for reviewing my PR @mnlevy1981. I've modified the clear.py script and README.md as you suggested (and also removed redundant exceptions for the ones I noticed python reports). Please let me know how this looks.
> 
> Thanks for making those changes! Could you add the `try` / `except` block that you removed from `readConfigFile()` to `get_control_dict()`?

Sure, I just made that change.

TeaganKing:
I do see the benefit of adding that try/except block to [get_control_dict()](https://github.com/shivaniikum/CUPiD/blob/0cb7d6d6454be0367a36255fecd73c7ea6831bdd/cupid/util.py#L76C5-L76C21) to clarify the file not found error and avoid clutter, since this error may come up somewhat often.

Regarding the `run_dir` error checking, I don't have a strong preference between using a sys.exit and raising a ValueError. It would reduce clutter to use sys.exit, but I think it's still fairly clear that there's a file not found error. My only other concern with skipping the tracebacks would be that if errors occur for unexpected reasons, it could be harder to debug. So, either way is fine by me.

shivaniikum:
Thanks for clarifying. I just added the changes you suggested. I agree that these changes make the errors easier to understand.

shivaniikum:
I agree that these changes make the errors easier to understand. I've modified util.py and clear.py as suggested by @mnlevy1981.

rmshkv:
I think I'd vote for the ValueError rather than sys.exit, looks a little less clean but might be better to keep things consistent.

rmshkv (approved):
This worked successfully for me! Besides clearing up the last thing Mike mentioned, I think it's good to merge. Thanks so much for your contributions and quick changes @shivaniikum!

cupid/clear.py
```diff
@@ -0,0 +1,47 @@
...
+            if os.path.isfile(itemPath):
+                os.remove(itemPath)  
+            # If item is a directory, recursively clear it
+            elif os.path.isdir(itemPath):
+                clearFolder(itemPath)   
+        # After deleting all items, remove the folder itself
+        os.rmdir(folderPath)    
+        print(f"All contents in {folderPath} have been cleared.")
+    except Exception as e:
+        print(f"Error occurred while clearing contents in {folderPath}: {e}")
```
mnlevy1981, line 23: Is this function necessary? You can use [`shutil.rmtree()`](https://docs.python.org/3/library/shutil.html#shutil.rmtree) to remove an entire directory (and its contents).

```diff
-import os
+import shutil
def clear(config_path):
    run_dir = readConfigFile(config_path)
-    clearFolder(run_dir)
+    shutil.rmtree(run_dir)
+    print(f"All contents in {run_dir} have been cleared.")
```

shivaniikum: Thank you for pointing that out. I wasn't aware of the rmtree function. I think rmtree is definitely better than the clearFolder function. I just modified clear.py to use rmtree.

cupid/clear.py
```diff
@@ -0,0 +1,32 @@
...
+    run_dir = control['data_sources'].get('run_dir', None)
+    
+    if run_dir:
+        #Append '/computed_notebooks' to the run_dir value if it is not empty
+        fullPath = os.path.join(run_dir, 'computed_notebooks')
+        return fullPath
+    
+    else: #run_dir is empty/wasn't found in config file so return error
+        print("ERROR: 'run_dir' was empty/not found in the config file.")
+        sys.exit(1)
```
mnlevy1981, line 23: I think @TeaganKing and @rmshkv would both prefer this specific `print` / `exit` to be a `raise ValueError` instead (while keeping the `sys.exit()` in `get_control_dict()`), and I'm okay with going back to that.

cupid/clear.py
```diff
@@ -0,0 +1,32 @@
...
+
+@click.command()
+@click.argument('config_path')
+#Entry point to this script
+def clear(config_path):
+    run_dir = readConfigFile(config_path)
```
mnlevy1981, line 29: One last minor thing: could you add a docstring to this function so that `cupid-clear --help` describes with the function does?

shivaniikum: I've added both changes. Please let me know how this looks

mnlevy1981 (approved):
This looks great!

shivaniikum:
That's awesome! Thank you so much for reviewing my PR and for all the suggestions @mnlevy1981 @TeaganKing @rmshkv!

TeaganKing:
Great work @shivaniikum ! Thank you for contributing this useful feature to the repository!

Issue #89, opened 2024-03-20
Ensure CUPiD is generalizable to the point where it can be run on other machines


TeaganKing:
To do this, we should remove references to `/glade` from code (eg, just have paths in config file), and test on other machines (eg, TACC).

TeaganKing:
We'll probably want to create a root directory variable in our global `config.yaml` file, and make relative file paths instead of specific ones like the following (thanks to @mnlevy1981 for this list):
```
diag_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/ADF-data/regrid/${diag_cam_climo.cam_case_name…
climo_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/
ts_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/
obs_data_loc: /glade/work/nusbaume/SE_projects/model_diagnostics/ADF_obs
cam_hist_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/${diag_cam_cli…
cam_tem_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/ADF-data/TEM/${diag_cam_climo.cam_case_…
cam_hist_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/${diag_cam_bas…
cam_climo_loc: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/ADF-data/climos/${diag_cam_baseline_c…
```

shivaniikum:
I am interested in working on this issue.

samsrabin:
It would probably be a good idea as well to avoid assuming *nix systems—I've seen some code where `/paths/are/assumed/to/look/like/this`, instead of `os.path.join("writing", "os-independent", "paths")`.

ekluzek:
Another area is use of /glade/derecho/scratch. Such as here:

``` python
helper_scripts/generate_ldf_config_file.py:        "/glade/derecho/scratch",
helper_scripts/generate_ldf_config_file.py:        "/glade/derecho/scratch",
helper_scripts/generate_ldf_config_file.py:        "/glade/derecho/scratch",
helper_scripts/generate_ldf_config_file.py:        "/glade/derecho/scratch",
helper_scripts/generate_ldf_config_file.py:        "/glade/derecho/scratch",
[...]

PR #92, opened 2024-04-10
PR for Issue #89: "Ensure CUPiD is generalizable and portable to other machines"

shivaniikum:
This pull request addresses issue #89 and contains changes to the config_f.cam6_3_119.FLTHIST_ne30.r328_gamma0.33_soae.001.yaml file at path "examples/nblibrary/config_f.cam6_3_119.FLTHIST_ne30.r328_gamma0.33_soae.001.yaml". 

I've removed hardcoded references to glade/ and added global variables to the top of the file and changed to relative file paths throughout the rest of the file. (Note: users should change the start and end year variables to match their cesm output files).


TeaganKing:
Hi @shivaniikum , thanks for getting this PR started, and nice work finding all of these `glade` paths.

One main thing that I'm noticing is that we want to generalize how we read in files while also maintaining the functionality of pointing to the current paths.

So, I think it would be useful to set a root directory in a `config.yml` file parallel to the `examples/coupled_model` directory (rather than the one in `nblibrary`), and use this to determine the full paths in `config_f.cam6_3_119.FLTHIST_ne30.r328_gamma0.33_soae.001.yaml` For instance, the `root_directory` variable could be something like `../local_cesm_output/` in your config.yml file, and something like `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/` in the `examples/coupled_model` config file.

@mnlevy1981 do you want to add anything else to this suggestion?

mnlevy1981:
Nothing to add to your specific suggestion, but I also wanted to ask @shivaniikum to merge the latest `main` into this branch since it was started before #91 was merged it

mnlevy1981:
Actually, one thing to add - we're okay with modifying the directory structure under `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/` - so if setting a root directory in `examples/coupled_model/config.yml` results in slightly different paths being passed to ADF that's fine. At some point in the future we'll decide on a file structure assuming the root directory is a path generated by CESM, and then we'll make sure any existing examples fit, but until then the directory structure we talked about last week is preferable to what currently exists on the NCAR machine.

mnlevy1981:
Two more requests before digging in for an actual review:

1. It looks like your `config.yml` file that gets data from `examples/local_cesm_output/` hasn't been added to this PR yet (if the decision is to make it `examples/local_cesm_output/config.yml`, then your root directory could be `./` instead of `../local_cesm_output`)
2. Could you please provide a small `README` in addition to `config.yml` describing how to get the necessary data from the web?

PR #100, opened 2024-05-17
Add pre-commit to cupid-dev

mnlevy1981:
I think we want to encourage the workflow of

1. checking out `CUPiD`
2. creating `cupid-dev` and `cupid-analysis`
3. activating `cupid-dev` and running `pre-commit install`
4. always running `git commit` from `(cupid-dev)`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,6 +1,6 @@
 ### All Submissions:
 
-* [ ] Have you followed the guidelines in our Contributing document (including the `pre-commit` check)?
+* [ ] Have you followed the guidelines in our Contributer's Guide](https://github.com/NCAR/CUPiD/wiki/Contributor's-Guide) (including the `pre-commit` check)?
 * [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
 
 <!-- You can erase any parts of this template not applicable to your Pull Request. -->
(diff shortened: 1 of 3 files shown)


TeaganKing:
Ah, I just saw these updates are also in the contributor's guide wiki page. Were you planning on adding anything else here? Maybe if we do want people to run `pre-commit install`, we should add that to the readme right before the `Running` section, perhaps in a section titled `set up pre-commit`?

mnlevy1981:
> Ah, I just saw these updates are also in the contributor's guide wiki page. Were you planning on adding anything else here? Maybe if we do want people to run `pre-commit install`, we should add that to the readme right before the `Running` section, perhaps in a section titled `set up pre-commit`?

This is a good question - we don't want to give the impression that `pre-commit install` is required for running CUPiD, but do want users to run the command if they plan on doing any development... I think the README is currently geared towards users rather than developers, do we want to make it more general? Or point to the wiki? Or move the wiki page into a different markdown file that also ends up on the webpage?

TeaganKing:
> > Ah, I just saw these updates are also in the contributor's guide wiki page. Were you planning on adding anything else here? Maybe if we do want people to run `pre-commit install`, we should add that to the readme right before the `Running` section, perhaps in a section titled `set up pre-commit`?
> 
> This is a good question - we don't want to give the impression that `pre-commit install` is required for running CUPiD, but do want users to run the command if they plan on doing any development... I think the README is currently geared towards users rather than developers, do we want to make it more general? Or point to the wiki? Or move the wiki page into a different markdown file that also ends up on the webpage?

Ok, keeping the README as focused on users instead of developers makes sense to me. Can we maybe just add a link to the wiki page in the README and mention it maybe at the top of the installation header?

TeaganKing (approved):
Looks great! I just added the contributor's guide link to the pull request template.

PR #101, opened 2024-05-23
Minor update to NCAR_tips.md to remove config.yml

TeaganKing:
I just noticed we didn't update the NCAR tips README when we changed the formatting for how to run cupid with `cupid-run`


mnlevy1981 (approved):
Good catch!

PR #104, opened 2024-06-05
Add new example for generating key metrics

mnlevy1981:
The goal of this config file is to specify a case and compute a handful of key metrics as defined by in the CESM project meeting. (It may be the case that these key metrics require comparing to other CESM runs or observations.)


TeaganKing (approved):
Changed case name

PR #117, opened 2024-07-16
Add ability to specify generated time series file output path

### LINKED ISSUE #116: Need ability for user to select where generated time series are written
**Is your feature request related to a problem? Please describe.**

Currently the time series generation in CUPiD is hardcoded to write to the same directory path that the history files exist in.  However, there are situations (like the CESM tutorial) where this may not be possible, as a user may only have read access to the original model run or archive directory.

**Describe the solution you'd like**
Can fix this issue by adding a new, optional `ts_output_dir` config YAML variable to the `timeseries` config group, which if present will override the default time series path with a new path provided by the user.

**Describe alternatives you've considered**
I imagine a command line argument could provide the same functionality, but it looks like all of them are logicals, so I figured adding an optional YAML variable would be better.

**Additional context**
I will open a PR for this shortly.

nusbaume:
Fixes #116 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/cupid/run.py
+++ b/cupid/run.py
@@ -117,7 +117,7 @@ def run(
                     os.path.join(
                         global_params["CESM_output_dir"],
                         timeseries_params["case_name"],
-                        f"/{component}/proc/tseries/",
+                        f"{component}", "proc", "tseries",
                     ),
                 ]
 
@@ -125,7 +125,7 @@ def run(
                     ts_output_dir = [
                         os.path.join(
                             timeseries_params["ts_output_dir"],
-                            f"{component}/proc/tseries/",
+                            f"{component}", "proc", "tseries",
                         ),
                     ]
                 # -----
--- a/cupid/run.py
+++ b/cupid/run.py
@@ -113,21 +113,17 @@ def run(
 
                 # set time series output directory:
                 # -----
-                ts_output_dir = [
-                    os.path.join(
-                        global_params["CESM_output_dir"],
-                        timeseries_params["case_name"],
-                        f"{component}", "proc", "tseries",
-                    ),
-                ]
-
                 if "ts_output_dir" in timeseries_params:
-                    ts_output_dir = [
-                        os.path.join(
+                    ts_output_dir = os.path.join(
                             timeseries_params["ts_output_dir"],
                             f"{component}", "proc", "tseries",
-                        ),
-                    ]
+                    )
+                else:
+                    ts_output_dir = os.path.join(
+                            global_params["CESM_output_dir"],
+                            timeseries_params["case_name"],
+                            f"{component}", "proc", "tseries",
+                    )
                 # -----
 
                 # fmt: off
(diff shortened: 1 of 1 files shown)


nusbaume:
@mnlevy1981 @TeaganKing please let me know what tests you want to have run with this branch as specified in the checklist above (I obviously tested it in my own config file).  Thanks!

TeaganKing:
Hi @nusbaume , We don't have robust testing set up yet (that PR template checklist was a bit preemptive), so I think just testing with your own config file is great.

TeaganKing:
It looks like the pre-commit is also passing.  Whenever it's ready, feel free to request a review.

nusbaume:
@TeaganKing thanks for letting me know!  Sadly it looks like I am unable to add reviewers to this PR.

cupid/run.py
```diff
@@ -110,6 +110,26 @@ def run(
...
+                    ),
+                ]
+
+                if "ts_output_dir" in timeseries_params:
+                    ts_output_dir = [
+                        os.path.join(
+                            timeseries_params["ts_output_dir"],
+                            f"{component}", "proc", "tseries",
+                        ),
+                    ]
```
mnlevy1981, line 130: One suggestion: can we keep `ts_output_dir` as a string, and pass `[ts_output_dir]` to `cupid.timeseries.create_time_series()`? (A separate issue ticket might be to update `create_time_series` to accept strings or lists as arguments, so list-ifying a handful of arguments seems silly... but that will be easier to address if `ts_output_dir` is a string.)

Also, a question: an alternate approach to this block of code would be

```py3
                if "ts_output_dir" in timeseries_params:
                    ts_output_dir =os.path.join(
                        timeseries_params["ts_output_dir"],
                        f"{component}", "proc", "tseries",
                    )
                else:
                    ts_output_dir = os.path.join(
                        global_params["CESM_output_dir"],
                        timeseries_params["case_name"],
                        f"{component}", "proc", "tseries",
                    )
```

Is there a reason to prefer one implementation over the other? If you want to keep

```py3
var=value
if condition:
    var=other_value
```

I see one `if-else` block in `run.py`, could you change that one?

```diff
        # Doing initial subsetting on full catalog, e.g. to only use certain cases

+        cat_path = full_cat_path
        if "subset" in control["data_sources"]:
            first_subset_kwargs = control["data_sources"]["subset"]
            cat_subset = full_cat.search(**first_subset_kwargs)
            # This pulls out the name of the catalog from the path
            cat_subset_name = full_cat_path.split("/")[-1].split(".")[0] + "_subset"
            cat_subset.serialize(
                directory=temp_data_path, name=cat_subset_name, catalog_type="file",
            )
            cat_path = temp_data_path + "/" + cat_subset_name + ".json"
-        else:
-            cat_path = full_cat_path
```

mnlevy1981 (changes requested):
This looks great! Just one small request and then a quick question that will lead to a second small request regardless of your answer :)

cupid/run.py
nusbaume: I am happy with either method of writing the if-statement.  I personally prefer the `if-else` method, but have found that some static analysis tools like `pylint` will complain about it, hence I went the `var=value, if condition: var=other_value` way in this PR.  Do you all have a preference one way or the other?

mnlevy1981: Let's stick with `if-else`, and if we eventually add `pylint` or some other checker that complains we'll just have one more block to update.

nusbaume: Sounds good!  I've applied the requested changes.

mnlevy1981 (approved):
looks great!

Issue #93, opened 2024-04-10
Allow scripts to be run instead of notebooks


TeaganKing:
We'd like to be able to run a script rather than just a notebook with CUPiD.

A good test of this would be to run a sea ice script instead of a sea ice notebook.

This will require saving figures generated from the notebook and reading them in with a standard naming convention to an html file that could show the figures.  See #16 for the html component of this.

mnlevy1981:
@dabail10 started to work on this and ran into a major stumbling block -- the `ploomer` script runner doesn't offer an obvious way to change environments before running the script. Dave's example was being run in `cupid-dev`, which doesn't have the necessary analysis packages installed. Lev will dig into the ploomer documentation and then I'll work with Dave to get a fix in place.

TeaganKing:
Just for reference, the function which includes ScriptRunner is [create_ploomber_script_task](https://github.com/NCAR/CUPiD/blob/7134f3a8b88f2e1104be2ac80bbd0183c26b1ce8/cupid/util.py#L208C1-L209C1).

In the ScriptRunner documentation, it looks like we should be able to specify the environment in the configuration file using something like this, adapted for the script that you have:

```
tasks:
  - source: script.py
  - product: output.txt
  - env: cupid-analysis
```

So, specifying 'env' *should* ideally work to allow for using the `cupid-analysis` environment. Maybe we can test it out tomorrow?

rmshkv:
Hmm, maybe I missed something - could you point me to where in the documentation or Ploomber source code that feature is mentioned?

TeaganKing:
Apologies, I may have jumped ahead without enough research on this... I saw this in an example, but after another look, I'm now thinking that the example was not accurate and it might not actually work in the format that I commented in the code snippet above. @rmshkv did you already have a particular place in the Ploomber code you were looking at to update?

TeaganKing:
It looks like the `kernelspec_name` in NotebookRunner is set roughly [here](https://github.com/ploomber/ploomber/blob/345f0a06b1baca57b20855449cce298329d2047a/src/ploomber/tasks/notebook.py#L749). Perhaps we could add an environment to [this part of ScriptRunner](https://github.com/ploomber/ploomber/blob/345f0a06b1baca57b20855449cce298329d2047a/src/ploomber/tasks/notebook.py#L957)?

Or, if we prefer to do something from our own side, could we possibly try something like this from within the script? I'm not sure what the best method is, but wanted to toss out some ideas.
```
[...]

PR #118, opened 2024-07-18
Key metrics

mnlevy1981:
Put together a "key metrics" example that can be run quickly for a single CESM case to look at a small number of important diagnostics. The first pass at this adds a land-ice notebook, but more notebooks are coming.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


TeaganKing:
@mnlevy1981 would it be helpful for me to review and then bring in the land-ice notebook as is, or should we wait until additional notebooks are ready to bring in?

mnlevy1981:
I have a few more changes to make for this one (pulling several functions out into a `utils.py` module and some other clean-up), then we should get feedback from @Katetc and @gunterl . I'm also tempted to move from `netCDF4` calls to `xarray` to be more in line with other CUPiD notebooks.

One other thing to note: this notebook currently reads output from coupler history files, but I think Gunter said that work is in progress to get those fields into the CISM history files -- depending on the timeline for that, we should either pull in the notebook after it's cleaned up or wait until we can update the location of the history files to read.

mnlevy1981:
This is ready for review, though I think `utils.py` probably needs a bit of work. Do we want to move plotting routines to a new module and keep `utils.py` for computational routines? Do we want to reorganize `utils.py` so the routines called from the notebook are at the top? As far as output goes, the map looks like

SMB vs obs

and the time series looks like

SMB time series

Note that I added the years included in the climotology to the map title and the legend; to test the "compare against a base case" (which is turned off by default), I compared years 62-101 to years 43-82 of the same run and realized it was useful to have different labels for those different plots.

mnlevy1981:
(sorry for the assigning / unassigning, I thought I was requesting reviews)

TeaganKing:
Thanks for all your work on this @mnlevy1981 ! I'm aiming to look at it tomorrow!

TeaganKing:
A few comments:
- Should these utils all necessarily be in the `glc` directory or is another more general location appropriate?
- As we discussed today, converting `netcdf4` --> `xarray` calls could also be useful.
- I can do some more thorough testing later but figured it may be worthwhile to make these larger comments first

mnlevy1981:
> * Should these utils all necessarily be in the `glc` directory or is another more general location appropriate?

I think we should keep them in `glc` until we have an example from another component that needs the same functions (or a generalized form that can be used both for land ice and the second component with different arguments). I'm not sure where the common utilities would go, though... my first thought was `cupid/` but that directory is for `cupid-dev` rather than `cupid-analysis`.

> * As we discussed today, converting `netcdf4` --> `xarray` calls could also be useful.

I'll work on this, and will probably end up with some of the functions we talked about (like a `cupid_open_dataset` that wraps `xr.open_dataset` or `xr.open_mfdatasets`). At which point the first bullet is worth revisiting, because that certainly wouldn't belong in the `glc/` directory :)

> * I can do some more thorough testing later but figured it may be worthwhile to make these larger comments first

Thanks! Verifying you can run the notebook on my branch would be great, but given the changes coming to address your first two points I don't think you need to dive too deep at this point

mnlevy1981:
As of 5b6e3d5 I think this is really ready for review. I've gotten rid of the `netcdf4` imports - we use `xarray` to open the datasets and then `da.sum()` and `da.mean()` instead of their numpy equivalents.

examples/nblibrary/glc/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    ax.set_xlabel("")
+    ax.set_ylabel("")
+    ax.set_xticklabels("")
+    ax.set_yticklabels("")
+    ax.set_xticks([])
+    ax.set_yticks([])
+
+
+def plot_contour(da, fig, ax, title, vmin, vmax, cmap, mm_to_Gt):
+    avg_data = np.round(da.sum().data * mm_to_Gt, 2)
```
TeaganKing, line 71: I guess it's fairly self explanatory but can you add descriptions of the functions and parameters here and on the plot_line function?

TeaganKing (commented):
I was able to run this successfully and replicate the plots that you showed.

One other somewhat odd feature is that the output is located in "CUPiD/examples/key_metrics/computed_notebooks/key_metrics"-- is there a reason for this second 'key_metrics' directory?

examples/nblibrary/glc/utils.py
mnlevy1981: Good catch, I'll add docstrings to `plot_contour()` and `plot_line()`!

mnlevy1981:
> One other somewhat odd feature is that the output is located in "CUPiD/examples/key_metrics/computed_notebooks/key_metrics"-- is there a reason for this second 'key_metrics' directory?

I thought we had an issue ticket about this, but I can't find it. I'd really like to clean up the names of the directories `cupid-run` creates -- maybe `CUPiD_out/` with `computed_notebooks/`, `computed_scripts/` (once we finish #93), and `html/` (created by `cupid-build` rather than `cupid-run`).

TeaganKing (approved):

PR #122, opened 2024-08-13
Some clean-up for key_metrics

mnlevy1981:
1. Standardize the way dates are passed (glacier notebook pulls year out of full date string to match atmosphere)
2. Add option for baseline comparison to atmosphere NMSE
3. Update config.yml to use run 104 (keeping 092 as baseline)
4. Clean up time series generation (allow empty list to specify "don't convert any files to time series for this component")

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


mnlevy1981:
There's not much change in the NMSE of surface pressure between 92 and 104:

439db59960de6dd183d7bc35c1f8c6ae294e6ceaa15ec20b14441c3a005a80d0

The surface mass balance is very different, likely due to (a) only averaging over 20 years instead of 40, and (b) only using 25 years for the spin-up instead of 60:

c133d0739aea764f4ac25e4f37c1a29ac106e5658bdc8a49b9fbe3dc431ed1aa

e6a1cf7998dad4b5061aae9d06572712a07fb4bc1d9e6c9f29e6791336fa1ff4

The x-axis of this plot could be extended -- there's 40 years of data from 092 but only the last 20 years are shown 

664ca78d05d5241363c25efd1776f9d251cb72f32a38a8a2e92c7dcdcff9c718

cupid/timeseries.py
```diff
@@ -79,6 +79,11 @@ def create_time_series(
 
     """
 
+    # Don't do anything if list of requested diagnostics is empty
```
TeaganKing, line 82: Thanks for adding this!

examples/key_metrics/config.yml
```diff
@@ -42,55 +42,59 @@ computation_config:
...
 
   ocn:
-    vars: ['taux', 'tauy'] # ['process_all']
+    vars: []
     derive_vars: []
-    hist_str: 'h.frc'
+    hist_str: 'h.z'
     start_years: [1]
-    end_years: [100]
+    end_years: [44]
```
TeaganKing, line 81: Should we determine this from the start/end date?

cupid/timeseries.py
mnlevy1981: You're welcome! It feels a little bit like a hack, though - what I'd really like to do is set it up so that if you don't want ocean time series files to be generated you just remove the `ocn` block from the YAML file; that currently aborts with

```py3
  File "/glade/work/mlevy/conda-envs/cupid-dev/bin/cupid-run", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/glade/work/mlevy/conda-envs/cupid-dev/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/glade/work/mlevy/conda-envs/cupid-dev/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/glade/work/mlevy/conda-envs/cupid-dev/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/glade/work/mlevy/conda-envs/cupid-dev/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/glade/work/mlevy/codes/CUPiD/cupid/run.py", line 133, in run
    timeseries_params[component]["vars"],
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'ocn'
```

I think we just need to treat the loop through `component` for time series the same way we do for running notebooks, where we continue to the next component if there isn't a block for a given component (maybe we throw an error if that component was explicitly requested? `cupid-run --time-series --ocean` probably should still fail if there isn't a `timeseries.ocn` block in the config file; I think that's what happens with `cupid-run --ocean`)

examples/key_metrics/config.yml
mnlevy1981: The way we have things set up, we could do

```yaml
start_years: [1, 11, 21, 31, 41]
end_years: [10, 20, 30, 40, 44]
```

And get decade-long time series, so we don't necessarily want these always be determined from the start and end dates. That said:

1. I don't think we need to have per-component `start_years` and `end_years`, and
2. Maybe we could add something like `max_years_per_timeseries_file` to the yaml file and then generate `start_years` and `end_years` from that value combined with `start_date` and `end_date`

TeaganKing: Right. Okay, perhaps an explicit list of years specified for the timeseries is best then. I could see `max_years_per_timeseries_file` being useful, but I do think it may start to obscure what's actually happening from the user.

cupid/timeseries.py
mnlevy1981: I have a fix for this in my branch used for regridding CESM output, but I'll this branch as-is and then get the cleaner fix in with the regridding PR (which doesn't exist yet)

examples/key_metrics/config.yml
mnlevy1981: It turns out that `start_years` and `end_years` are lists, but the length is meant to match the length of `case_name` value... so

```yaml
start_years: [1, 11, 21, 31, 41]
end_years: [10, 20, 30, 40, 44]
```

uses years 1-10 for the first case, 11-20 for the second case, etc... if there is only one case, then the second value onward is ignored. I don't think the current time series generation tool really offers a clean way specify a single range of years but get multiple time series files spanning that range.

Should `max_years_per_timeseries_file` come in with this PR, or should it be something separate?

mnlevy1981:
Still to-do:

- [ ] Resolve `max_years_per_timeseries_file` discussion
- [ ] Update 104 to include 100 years of output, for more apples-to-apples comparison

examples/key_metrics/config.yml
TeaganKing: I think `max_years_per_timeseries_file` could be something separate if we do decide to implement it, just since it seems a bit less clear than the current implementation.

mnlevy1981:
Updated figures comparing 100 years of each run:

**NMSE**

fdb7a2c0abedc4ea96c9ed7808b768771b74b8d8cc81354f5a743a9c4376bfb7

**SMB**

bd9de61aff4798bc03bc0aef11bb46042e35a29991bc0ab785edcf7c8192bc44
b3d391f5abd5aa12e95535aa59bba6cc3c022b7f8cc0d6300f6ef1ad1f2fd0fb
767cc7e08d4e14b4dfaf8235a67fdaa633581b30874354d2ac2e28c00688b564

I really like how the 104 vs 92 SMB plot has a sign change just above the southern tip of the island, that seems pretty interesting. Anyway, this is ready for final review / merging :)

mnlevy1981:
It's interesting that the black line in

time series

doesn't extend as far as the red and blue lines... going to try to fix that real quick

mnlevy1981:
> It's interesting that the black line in
> 
> time series
> 
> doesn't extend as far as the red and blue lines... going to try to fix that real quick

fixed in c0ab02a

109c424112e4a084600f18d6cab6c69029b07bb5ce79eeacd1689bccbb4662f9

TeaganKing:
@mnlevy1981 is this ready for review?

mnlevy1981:
Yup, I think this is ready to go. Latest output is in https://github.com/NCAR/CUPiD/pull/122#issuecomment-2316349738 (except for the time series plot, which is updated in https://github.com/NCAR/CUPiD/pull/122#issuecomment-2316361593)

TeaganKing (commented):
Thanks for doing all of this clean up! The method for passing dates is looking much clearer, and the capability of adding an empty list is helpful. The plot is looking a lot better, too! I'm going to test this out locally as well.

TeaganKing (approved):
Looks good to me! Local testing worked smoothly. All tasks are completed despite the tasklist checker failing... I think this is because the last checkbox was checked off after the last commit was made.

Issue #82, opened 2024-03-12
Add logging capabilitites


TeaganKing:
We should replace print statements with logging. It would be good to include various levels of logging, as well.

PR #127, opened 2024-08-29
[WIP] Logging

TeaganKing:
This PR should address #82 by adding logging capabilities to CUPiD instead of using print statements. This should accomplish the following three tasks:

- [x] update print statements to logger statements
- [x] add log file to `.gitignore`
- [x] move ploomber output to log file instead of printed to console (moved to new issue #129)

It is useful to implement logging instead of using print statements because log files can be retrieved in the future (whereas print statements are not saved) and this reduces the amount of printing to the console that can get overwhelming for users.

### All Submissions:

### New Feature Submissions:

1. [x] Have you lint your code locally prior to submission?

### Changes to Core Features:


TeaganKing:
[This page](https://docs.ploomber.io/en/latest/user-guide/logging.html) provides some examples of basic pipelines that can be used to set up logging for ploomber; however, this calls dag in a slightly different manner and will require a bit more looking into.

TeaganKing:
Given that within the CESM workflow, log files such as `cupid.o####` and `cupid.e####` where `####` is a timestamp will be automatically generated, we can remove the filename from the logger instantiation.

- [ ] It would be useful to set up logging levels in the configuration file so that users can choose what amount of output they want printed to the console.

TeaganKing:
I'd like to test this a bit more thoroughly to make sure it didn't break cupid-run or timeseries generation. Thus, the testing might be on hold until HPC systems are back up, but this should be ready for testing/review next week.

cupid/util.py
```diff
@@ -93,6 +95,39 @@ def get_control_dict(config_path):
...
+            )
+        if log_level == "error":
+            logging.basicConfig(
+                level=logging.ERROR,
+            )
+    else:
+        # default level is info if log level is not set in config
+        logging.basicConfig(
+            level=logging.INFO,
+        )
```
mnlevy1981, line 126: If you change line 104 to

```py3
log_level = control["computation_config"].get("log_level", "info")
```

Then you don't need the `if` statement at 105 or the entire `else` block from 122 - 126 (maybe move the `# default level is info if log level is not set in config` comment to above setting `log_level`?)

cupid/timeseries.py
```diff
@@ -347,10 +351,12 @@ def create_time_series(
...
 
     # Notify user that script has ended:
-    print(f"  ... {component} time series file generation has finished successfully.")
+    logger.info(
+        f"  ... {component} time series file generation has finished successfully.",
+    )
 
 
-def derive_cam_variables(vars_to_derive=None, ts_dir=None, overwrite=None):
+def derive_cam_variables(vars_to_derive=None, ts_dir=None, overwrite=None, logger=None):
```
mnlevy1981, line 359: This function is called from `create_time_series()`, which uses `logger` prior to calling in... so I think we want `logger` to be a required argument. This will let us remove

```py3
if logger is None:
    logger = cupid.util.setup_logging("config.yml")
```

at lines 370-372 (and I think also the `import cupid.util` at lines 17 & 18)

mnlevy1981 (changes requested):
This looks great! There are a couple of in-line comments, plus I wanted to note that I grepped for `print` (only in the `cupid/` subdirectory) and found two more occurrences that we may want to address:

```bash
$ cd cupid
$ git grep -n print
timeseries.py:87:        print(f"\n  No time series files requested for {component}...")
util.py:55:        print(f"ERROR: {config_path} not found")
```

cupid/util.py
mnlevy1981: Thinking about this more, should we use

```py3
        if log_level == "debug":
            logging.basicConfig(
                level=logging.DEBUG,
            )
        elif log_level == "info":
            logging.basicConfig(
                level=logging.INFO,
            )
        elif log_level == "warning":
            logging.basicConfig(
                level=logging.WARNING,
            )
        elif log_level == "error":
            logging.basicConfig(
                level=logging.ERROR,
            )
        else:
```

and trap the case when `log_level` is not one of the four expected strings?

TeaganKing:
Thanks for this review, @mnlevy1981 !

I addressed those comments, except for the one print statement in `cupid/util.py`-- the get_control_dict function is needed _prior_ to logging implementation (because logging uses info from the config file). Thus, I think this one would have circular implementation if both functions call one another.

cupid/util.py
```diff
@@ -93,6 +96,34 @@ def get_control_dict(config_path):
...
+    if log_level == "warning":
+        logging.basicConfig(
+            level=logging.WARNING,
+        )
+    if log_level == "error":
+        logging.basicConfig(
+            level=logging.ERROR,
+        )
+
```
mnlevy1981, line 123: Did you see my follow-up comment about this (https://github.com/NCAR/CUPiD/pull/127#discussion_r1767067652)? What happens if someone puts `log_level: all_messages` (or even just `log_level: DEBUG`) in `config.yml`? none of these if statements will evaluate to true, so we won't call `logging.basicConfig()` at all... I think we want to catch that and raise an exception.

TeaganKing: Oh! I think I marked it as resolved and then the response was hidden. Sorry about that.

I added an 'else' statement back in and also included the all-caps versions of the log levels.

mnlevy1981 (approved):
This looks great, thanks!

PR #131, opened 2024-09-07
Support cesm workflow

mnlevy1981:
Code changes that will be useful when running CUPiD via the CESM workflow.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


mnlevy1981:
I added the `generate_cupid_config_file.py` that @ingridc2051 wrote to CUPiD, so that `template.cupid` in https://github.com/ESMCI/ccs_config_cesm/pull/176 can find it. I made a couple small changes:

1. instead of importing some CIME functions, it expects CESM to pass command line arguments (`$CASE`, `$DOUT_S_ROOT`, and the location of CUPiD based off of `$SRCROOT`).
2. I added the `sort_keys=False` argument to `yaml.dump` so the resulting yaml file looks similar to the template (although missing the comments)

I'm not sure I like the location of the script, which is currently a new `cesm_scripts/` subdirectory. I also think we need something more information than `# sample comment` in the file header... I can update that to something like `# This file has been auto-generated for use with {{case}} based off of {{template file}}` but I'm open to suggestions

TeaganKing:
1. Can you clarify your reasoning for using command line arguments instead of importing CIME functions? I had originally thought the CIME functions were a bit more streamlined.
2. If we aren't going to be adding additional cesm scripts, I wonder if this could just go in the `cupid` directory?
3. I think that suggestion to update the comment is great.

mnlevy1981:
> 1. Can you clarify your reasoning for using command line arguments instead of importing CIME functions? I had originally thought the CIME functions were a bit more streamlined.

The original version of this code was placed in the CIME source tree, so it could easily import some CIME functions:

```py3
from standard_script_setup import *
from CIME.case import Case
```

Now that this script is owned by CUPiD, it would need a `sys.path.append()` call in front of those import statements, which would require a command line argument to pass in the path to these packages (or at least `${SRCROOT}`). Also, we eventually want the user to be able to specify which CUPiD example should provide the base `config.yml` file.

I do think we could replace `--cupid-root` with `--cesm-root` (just adding an additional `tools/CUPiD` to the path), and then we could also import the CIME libraries and get both `--case` and `--cesm-output-dir` from the `Case` object. I'll work on that later this week (or maybe next week, a lot of this week is taken up by a workshop).

> 2. If we aren't going to be adding additional cesm scripts, I wonder if this could just go in the `cupid` directory?

I think we want to keep it separate, especially if we are going to pass `--cesm-root` as an argument. Also, the `cupid/` directory should be reserved for key parts of the `(cupid-dev)` environment, while this script doesn't really require `cupid-dev` (it just needs to be run in an environment with `pyyaml`). Maybe the directory should be `helper_scripts/` and this could be renamed `generate_cupid_config_for_cesm_case.py`?

> 3. I think that suggestion to update the comment is great.

👍

TeaganKing:
Got it, that makes sense. The renaming looks good, too. I'll let you work on that replacement, and then can you please ping me again for review when it's ready?

mnlevy1981:
> Got it, that makes sense. The renaming looks good, too. I'll let you work on that replacement, and then can you please ping me again for review when it's ready?

This one is ready for review (note the default directory for `--case-root` is whatever directory you run the script from, which I denote as `{$PWD}` but will actually appear in the help message as the full path):

```
$ ./generate_cupid_config_for_cesm_case.py -h
usage: generate_cupid_config_for_cesm_case.py [-h] [--cesm-root CESM_ROOT] [--cupid-example CUPID_EXAMPLE]
                                              [--case-root CASE_ROOT]

Generate config.yml based on an existing CUPID example

options:
  -h, --help            show this help message and exit
  --cesm-root CESM_ROOT
                        Location of CESM source code (default: None)
  --cupid-example CUPID_EXAMPLE
                        CUPiD example to use as template for config.yml (default: key_metrics)
  --case-root CASE_ROOT
                        CESM case directory (default: {$PWD})
```

helper_scripts/generate_cupid_config_for_cesm_case.py
```diff
@@ -0,0 +1,77 @@
...
+        help="Location of CESM source code",
+    )
+
+    # Command line argument for CUPiD example from which to get config.yml
+    parser.add_argument(
+        "--cupid-example",
+        action="store",
+        dest="cupid_example",
+        default="key_metrics",
+        help="CUPiD example to use as template for config.yml",
```
TeaganKing, line 33: The default as key_metrics is great. Is it worthwhile to mention that the example choices are coupled_model and key_metrics? I'm not sure how many different examples we'll end up having?

helper_scripts/generate_cupid_config_for_cesm_case.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    # Command line argument location of CESM case directory
+    parser.add_argument(
+        "--case-root",
+        action="store",
+        dest="case_root",
+        default=os.getcwd(),
```
TeaganKing, line 41: It may also be worth logging information on what arguments were used?

TeaganKing:
Just a few minor comments, and otherwise this looks good to me!

helper_scripts/generate_cupid_config_for_cesm_case.py
```diff
@@ -0,0 +1,77 @@
...
+    with open("config.yml", "w") as f:
+        # write a comment
+        f.write(f"# This file has been auto-generated for use with {case}\n")
+        f.write(f"# It is based off of examples/{cupid_example}/config.yml\n")
+        # enter in each element of the dictionary into the new file
+        yaml.dump(my_dict, f, sort_keys=False)
+
+
+if __name__ == "__main__":
+    args = vars(_parse_args())
```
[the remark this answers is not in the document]
mnlevy1981: I added `required=True` for `--cesm-root` and now running with no arguments returns

```bash
$ ./generate_cupid_config_for_cesm_case.py
usage: generate_cupid_config_for_cesm_case.py [-h] --cesm-root CESM_ROOT [--cupid-example CUPID_EXAMPLE]
                                              [--case-root CASE_ROOT]
generate_cupid_config_for_cesm_case.py: error: the following arguments are required: --cesm-root
```

I think the default values for `--cupid-example` and `--case-root` are okay, though I can make `--case-root` required if you would prefer that.

helper_scripts/generate_cupid_config_for_cesm_case.py
mnlevy1981: This is a tough one -- I'd love to use the `choices` argument of `add_argument()` along with the `os.walk()` function to list the subdirectories of `CUPiD/examples/`, but we don't know where the `examples/` directory is until we resolve `--cesm-root`. The best option is probably to enforce this immediately after entering `generate_cupid_config()` by ensuring that `os.path.join(cupid_root, "examples", cupid_example)` is a directory and printing out a useful error message if it isn't?

TeaganKing: That makes sense that adding the `choices` would entail using information we don't have yet. The suggestion you shared sounds great!

helper_scripts/generate_cupid_config_for_cesm_case.py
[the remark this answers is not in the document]
TeaganKing: I think this is great! The default for case-root seems fine to me.

helper_scripts/generate_cupid_config_for_cesm_case.py
mnlevy1981: The code in bb3f37d has the check and gives the following:

```bash
$ ./generate_cupid_config_for_cesm_case.py --cesm-root ${CESM_ROOT} --case-root ${CESM_ROOT}/cases/test_cupid_workflow.007 --cupid-example foo
  File "${CESM_ROOT}/tools/CUPiD/helper_scripts/./generate_cupid_config_for_cesm_case.py", line 107, in <module>
    generate_cupid_config(**args)
  File "${CESM_ROOT}/tools/CUPiD/helper_scripts/./generate_cupid_config_for_cesm_case.py", line 62, in generate_cupid_config
    raise KeyError(
KeyError: "--cupid-example must be a subdirectory of ${CESM_ROOT}/tools/CUPiD/examples (['coupled_model', 'key_metrics'])"
```

Note that both `ilamb` and `nblibrary` are left off the list of `valid_examples`; eventually we'll probably reinstate `ilamb`, but for now it doesn't have a `config.yml` file. I'd really like to move `nblibrary/` out of `examples/` but that seems like a big thing to tackle in this PR :)

I should probably add text about `'foo' is not a valid option for `--cupid-examples`; doing that now.

TeaganKing: Sounds good!

Yes, let's move nblibrary out of examples after this PR is merged in.

mnlevy1981: 0888657 has the improved check:

```bash
$ ./generate_cupid_config_for_cesm_case.py --cesm-root ${CESM_ROOT} --cupid-example foo
  File "${CESM_ROOT}/tools/CUPiD/helper_scripts/./generate_cupid_config_for_cesm_case.py", line 108, in <module>
    generate_cupid_config(**args)
  File "${CESM_ROOT}/tools/CUPiD/helper_scripts/./generate_cupid_config_for_cesm_case.py", line 63, in generate_cupid_config
    raise KeyError(
KeyError: "argument --cupid-example: invalid choice 'foo' (choose from subdirectories of ${CESM_ROOT}/tools/CUPiD/examples: ['coupled_model', 'key_metrics'])"
```

helper_scripts/generate_cupid_config_for_cesm_case.py
mnlevy1981: dba6bea adds information about the arguments to the comment section at the head of the new `config.yml`; I think that's the best place to put it but I'm open to other suggestions. (If we use a `print()` statement, then I think it will show up in the PBS log file generated by the `case.cupid` job, but other queue managers might treat that output differently?)

TeaganKing: This seems like a good place-- especially since the config.yml file also says it's generated by the script, so all that information is in one area.

TeaganKing (approved):
This looks great! Thanks for implementing this feature and making all those changes :)

PR #141, opened 2024-10-10
Create a new example that will run full ADF package (not from notebook)

mnlevy1981:
This example can be generalized to other outside diagnostic packages as well (e.g. ILAMB). The goal is to have the JupyterBook-generated web site link to the ADF website from the side bar.

### All Submissions:

* [x] Have you followed the guidelines in our [Contributor's Guide](https://github.com/NCAR/CUPiD/wiki/Contributor's-Guide) (including the `pre-commit` check)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change? Other pull requests that create the same changes do not exist; however, this is dependent upon #142 

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you lint your code locally prior to submission?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you successfully tested your changes locally?

PR #142, opened 2024-10-14
generate ADF config file

brianpm:
### New Feature Submissions:

3. [x] Have you lint your code locally prior to submission?

Not sure about tests... but yes, linted code locally.

### Changes to Core Features:

This adds a script that is a first attempt to take the CUPID config file to generate an ADF config file. The script works locally. 

The idea is to be able to run the ADF without the user needing to manually edit the ADF config file. 

I think we need to discuss what are the requirements for such an input file. It seems like ADF is expecting more information than CUPID right now... which is fine, but might mean that this script needs to make quite a few assumptions along the way.


brianpm:
I just pushed some updates (mostly untested). I think this gets us pretty close to a version that will run ADF.

This script would be run as (for example):
```python generate_adf_config_file.py --cupid_file ~/Code/CUPiD/examples/key_metrics/config.yml --adf_template ~/Code/CAM_diagnostics/config_amwg_default_plots.yaml --out_file ~/Desktop/testfile2.yaml```

Then, in principle, ADF would be run like:
```
> cd $ADFSOURCE
> ./run_adf_diag ~/Desktop/testfile2.yaml
```

Note that I am pretty sure the path for ADF output won't work yet. I set it to CUPiD's "sname", but I think it needs to be expanded to be a full path.

**Comment by TeaganKing** [comment]
I made a few minor formatting tweaks and additionally adjusted the path to ADF output based on #141 as well as updated the file paths to include the base or test case name after the CESM output dir.

I was able to create an ADF config file with a similar command as the one you shared above (just using my own files).

I do want to note that ADF does not run in the `cupid-dev` environment (which is reasonable based on expected uses for these environments), and thus requires activating `cupid-analysis` before running `run_adf_diag`.

ADF starts running with this configuration file, but does not seem to find all of the files needed for input. I think this is because `hist_str` is set to `cam.h0` and the adf lib expects to find data with the following format, but only finds "cam.h0i" files. We should probably update this, but I'm not sure if there's a consistent format we'd expect?
`/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/b.e23_alpha17f.BLT1850.ne30_t232.092/atm/hist/*cam.h0.*`

**Comment by TeaganKing** [comment]
Some updates on plot generation: the hanging after generating `ts` and `climo` files is related to processing the 3D variables. There also is a KeyError when looking for `lat` while generating zonal mean plots.

**Comment by TeaganKing** [comment]
The default config file for ADF has been updated (thanks to Justin!) and now the hist_str is just set to h0a to use averaged fields.

When running just PRECT and using an updated case for comparison (b.e30_beta02.BLT1850.ne30_t232.104 vs f.e23_alpha17f.FLTHIST_ne30.roughtopo.099), this runs fine and doesn't hang. It looks like that was a grid-related issue.

**Comment by TeaganKing** [comment]
A variable list could be added to the cupid config file that overwrites the default ADF variables.

**Comment by TeaganKing** [comment]
The `external_diag_packages` `config.yml` in #141 has been updated to include same-grid cases and a list of `adf_vars` which now can be used by this helper script (but is not necessary so that the script is compatible with other yaml files).

File: helper_scripts/generate_adf_config_file.py [hunk 1]
```diff
@@ -0,0 +1,187 @@
...
+        "--adf_template",
+        action="store",
+        required=True,
+        help="an adf config file to use as a base",
+    )
+    parser.add_argument(
+        "--out_file",
+        action="store",
+        required=True,
+        help="the output file to save",
```
mnlevy1981, line 34: Two comments on this, based on being consistent with `generate_cupid_config_for_cesm_case.py`:

1. The options should use`-` as a separator rather than `_` (`--cupid-file`, which will still be saved as `args.cupid_file`)
2. This script takes slightly different arguments than `generate_cupid_config_for_cesm_case.py`, wanting a full path to a CUPiD config file rather than finding a specific CUPiD example config file to modify. I feel fairly certain that we want to use the same mechanism for finding the CUPiD config file in both scripts, and am somewhat sure that providing the path to the file makes more sense than trying to construct if from other arguments... in which case this PR should modify the other script to match what is done here

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def generate_adf_config(cupid_file, adf_file, out_file):
+    """Use cupid_file (YAML) and adf_file (YAML) to produce out_file
+    by modifying adf_file with data from cupid_file.
+    """
+    with open(cupid_file, encoding="UTF-8") as c:
+        c_dict = yaml.load(c, Loader=yaml.SafeLoader)
+    with open(adf_file, encoding="UTF-8") as a:
+        a_dict = yaml.load(a, Loader=yaml.SafeLoader)
```
mnlevy1981, line 46: Does the `encoding` need to be specified here? And is there a difference between specifying `Loader=yaml.SafeLoader` in `yaml.load` and just using `yaml.safe_load`? The other file in this directory uses

```py3
    with open(os.path.join(cupid_root, "examples", cupid_example, "config.yml")) as f:
        my_dict = yaml.safe_load(f)
```

And I'm in favor of consistency whenever possible, whether we modify this script or `generate_cupid_config_for_cesm_case.py`

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+    base_case_name = c_dict["global_params"].get("base_case_name")
+    test_case_name = c_dict["global_params"]["case_name"]
+    c_ts = c_dict["timeseries"]
+    ts_case_names = c_ts.get("case_name")
+    if not ts_case_names:
+        raise ValueError("CUPiD file does not have timeseries case_name array.")
+
+    # Set case names for ADF config
+    a_dict["diag_cam_climo"]["cam_case_name"] = test_case_name
+    a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name
```
mnlevy1981, line 60: Is it okay for this value to be `None`? Do we need something like

```py3
if base_case_name:
  a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name
elif "cam_case_name" in "cam_case_name":
  del a_dict["diag_cam_baseline_climo"]["cam_case_name"]
```

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    start_date = get_date_from_ts(c_ts["atm"], "start_years", test_case_cupid_ts_index)
+    end_date = get_date_from_ts(c_ts["atm"], "end_years", test_case_cupid_ts_index)
+    a_dict["diag_cam_climo"]["start_year"] = start_date
+    a_dict["diag_cam_climo"]["end_year"] = end_date
+
+    # Set values for BASELINE
+    base_case_cupid_ts_index = (
+        ts_case_names.index(base_case_name) if base_case_name in ts_case_names else None
+    )
```
mnlevy1981, line 86: This block may as well be in the `if base_case_name is not None:` block

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+    if base_case_name is not None:
```
mnlevy1981, line 87: `if base_case_name:` would be consistent with the `ts_case_names` check above:

```py3
    ts_case_names = c_ts.get("case_name")
    if not ts_case_names:
        raise ValueError("CUPiD file does not have timeseries case_name array.")
```

mnlevy1981 (changes requested):
This looks great! I'm still putting together a test case to try it out myself, but it sounds like @TeaganKing has already done that so we are confident the script works as advertised :)

Two last consistency points:

1. `generate_cupid_config_for_cesm_case.py` is executable, and my inclination is that `generate_adf_config_file.py` should be as well (but I'm open to the idea of requiring `python generate_cupid_config_for_cesm_case.py` instead of allowing `./generate_cupid_config_for_cesm_case.py` if there is a good reason to avoid the latter)
2. The scripts in `cupid/` all use `click` instead of `argparse` and the scripts in this directory probably should as well. That's best handled in a separate PR, I just wanted to make a note here while I'm thinking about it.

helper_scripts/generate_adf_config_file.py
TeaganKing: It looks like these are the same functionally, the only difference is that `yaml.load` allows some extra flexibility but is slightly less readable if the only option included is `Loader=yaml.SafeLoader`. Thus, I changed this to `yaml.safe_load()` for consistency.

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+    a_dict["diag_cam_baseline_climo"]["cam_ts_loc"] = "/".join(
+        [base_case_output_dir, "atm", "proc", "tseries"],
+    )
+    a_dict["diag_cam_baseline_climo"]["cam_climo_loc"] = "/".join(
+        [base_case_output_dir, "atm", "proc", "climo"],
+    )
+    a_dict["diag_cam_baseline_climo"]["start_year"] = base_start_date
+    a_dict["diag_cam_baseline_climo"]["end_year"] = base_end_date
+
+    a_dict["diag_basic_info"]["hist_str"] = c_dict["timeseries"]["atm"].get("hist_str")
```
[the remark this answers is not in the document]
TeaganKing: I think if these are null they get set to start_date and end_date in lines 132 and 134

TeaganKing:
Thanks for these suggestions, @mnlevy1981 ! I think I have addressed everything (with a few notes in response to your direct in-line comments); let me know if you have any further requests.

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,187 @@
...
+    a_dict["user"] = "/".join(
+        [
+            "/".join(cupid_file.split("/")[:-1]),
+            "computed_notebooks",
+            c_dict["data_sources"]["sname"],
+            "_build",
+            "html",
+            "ADF",
+        ],
+    )
```
mnlevy1981: I just noticed these `"/".join()` calls that could be `os.path.join()` instead. Also, the `cupid_file.split("/")[:-1]` could be ` os.path.dirname(cupid_file)`

helper_scripts/generate_adf_config_file.py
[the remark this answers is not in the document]
mnlevy1981: I was thinking about line 119: `a_dict["diag_basic_info"]["hist_str"]  = c_dict["timeseries"]["atm"].get("hist_str")`; if `hist_str` is expected to always be in `c_dict["timeseries"]["atm"]`, then we should use `c_dict["timeseries"]["atm"]["hist_str"]` on the righthand side instead of the `.get()` function. If `hist_str` might not be available, then we should keep the `.get()` function but we probably don't want to set `a_dict["diag_basic_info"]["hist_str"] = None`

helper_scripts/generate_adf_config_file.py
mnlevy1981: Following up because I think the github formatting made it hard to see the context of my comment. This is specifically referring to `a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name`, since `base_case_name = c_dict["global_params"].get("base_case_name")` -- what is the expected behavior if `base_case_name` is not in `global_params`?

helper_scripts/generate_adf_config_file.py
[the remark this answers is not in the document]
TeaganKing: I think we'll be expecting `hist_str` to exist for numerous purposes in the config file, and it also shouldn't change all that often, so I think it'd be fine to use `c_dict["timeseries"]["atm"]["hist_str"]`.

helper_scripts/generate_adf_config_file.py
TeaganKing: This comment is based on an older version. I have removed the `cupid_file` (since we are now using `cupid_example`) and have already fixed the `"/".join()`.

helper_scripts/generate_adf_config_file.py
TeaganKing: In the ADF config file, `compare_obs` determines whether the model run is compared to observations or another model run. If this variable is false in the ADF config file, then `diag_cam_baseline_climo` is used and is a required config file section. If not present, ADF fails.

I'm curious if we would want to support comparison to obs in addition to model-model comparison-- @brianpm and @justin-richling do you have thoughts on this?

With all that said, I think that we could just use `base_case_name = c_dict["global_params"]["base_case_name"]` instead of the `.get()` since it is an expected value for model-model comparisons. I'll update this now.

helper_scripts/generate_adf_config_file.py
```diff
@@ -0,0 +1,265 @@
...
+    if not ts_case_names:
+        raise ValueError("CUPiD file does not have timeseries case_name array.")
+
+    # Set case names for ADF config
+    a_dict["diag_cam_climo"]["cam_case_name"] = test_case_name
+    if base_case_name:
+        a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name
+
+    elif "cam_case_name" in "cam_case_name":
+        del a_dict["diag_cam_baseline_climo"]["cam_case_name"]
```
mnlevy1981, line 90: Since you dropped the `.get()` for `base_case_name`, I think we want to drop the if/else and just keep line 87

mnlevy1981 (changes requested):
This looks great! I think all that's left is to hear @brianpm and @justin-richling's thoughts on https://github.com/NCAR/CUPiD/pull/142#discussion_r1846842721 and then update a few spots accordingly.

helper_scripts/generate_adf_config_file.py
justin-richling: I think we should allow for the model vs obs comparisons. In `adf_info.py` there is the variable `data_name` that gets set for the baseline case name if model vs model, but `data_name` gets set to "Obs" if its a model vs obs case.

helper_scripts/generate_adf_config_file.py
mnlevy1981: That's a reasonable concern! Maybe we can keep things as they are now, but update both `generate_adf_config_file.py` and `generate_cupid_config_for_cesm_case.py` in #141 if we want to drop the CESM dependency?

TeaganKing: That works for me!

helper_scripts/generate_adf_config_file.py
TeaganKing: Just updated this

helper_scripts/generate_adf_config_file.py
```diff
@@ -83,10 +83,9 @@ def generate_adf_config(cesm_root, cupid_example, adf_file, out_file):
 
     # Set case names for ADF config
     a_dict["diag_cam_climo"]["cam_case_name"] = test_case_name
-    if base_case_name:
-        a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name
+    a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name
 
-    elif "cam_case_name" in "cam_case_name":
+    if "cam_case_name" in "cam_case_name":
         del a_dict["diag_cam_baseline_climo"]["cam_case_name"]
```
mnlevy1981, line 89: I think we want to remove these two lines. The `if` statement isn't posed very well (it should probably be `if "cam_case_name" in a_dict["diag_cam_baseline_climo"]:`), but it was originally meant to catch the case where `a_dict["diag_cam_baseline_climo"]["cam_case_name"]` was read in from the template file and we wanted to remove it because `base_case_name` wasn't defined. Now we are always setting `a_dict["diag_cam_baseline_climo"]["cam_case_name"] = base_case_name` so we don't want to remove the dictionary entry.

TeaganKing: Done!

helper_scripts/generate_adf_config_file.py
TeaganKing: I'm a bit hesitant to include this feature within this PR if we want to get this in before Wednesday. I think it will require an additional default configuration file from ADF as well as a number of additional parameters in the cupid configuration file (including `compare_obs`) as well as parameters that are added to the script to generate the ADF config file.

TeaganKing: See issue #150

PR #143, opened 2024-10-15
Updated timeseries generation with GenTS 

AgentOxygen:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### Commentary
[GenTS](https://github.com/AgentOxygen/GenTS) is a modernized post-processing package that specializes in converting history files to timeseries files. All code changes are made to `run.py` as `timeseries.py` isn't needed (all of the functionality is encapsulated in GenTS). Release versions are made available via PyPI, so I opted to add it to the environment dependencies list rather than git fleximod or externals.

More testing is required for GenTS to make sure it is post-processing history files correctly and integrating into CUPiD will promote further testing. If CUPiD is nearing the production stage, then it might make more sense to create a separate branch.

GenTS can run in serial, but can run in parallel by utilizing Dask. Following the other notebooks, I create a local cluster unless serial is specified in the `config.yml`.

There are some timeseries specifications in the `config.yml` that don't yet exist in GenTS. For example, GenTS allows the user to specify a time slice, but this is generalized to all of the history files stored within a `ModelOutputDatabase` and cannot be broken into model components. To get around this, I create a unique `ModelOutputDatabase` for each model component, which is likely inefficient. GenTS does not require a history string to identify history files, but this comes with the caveat of processing *all* history files within the output directory. This may not be ideal for some use cases. I am unsure whether to implement these features into CUPiD or GenTS, as I would prefer to keep GenTS as generalized as possible but there may be some useful tools I could build into GenTS to enable this sort of configuration.

### DIFF
--- a/environments/cupid-analysis.yml
+++ b/environments/cupid-analysis.yml
@@ -34,6 +34,7 @@ dependencies:
   - ncar-jobqueue=2021.4.14
   - netcdf4=1.7.3
   - numpy=2.2.6
+  - pandas=2.3.3
   - pip=25.3
   - pluggy=1.6.0
   - pytest=9.0.2


cupid/run.py
```diff
@@ -83,110 +86,75 @@ def run(
...
-
     if time_series:
         timeseries_params = control["timeseries"]
+        client = None
+        if not global_params["serial"]:
+            cluster = LocalCluster(
+                n_workers=1,
+                processes=1,
+                threads_per_worker=1,
+                memory_limit="2GB",
```
mnlevy1981, line 107: Is there a reason to specify `memory_limit` rather than letting the `LocalCluster` object figure it out based on available resources? It seems like it could be problematic (if less than 2 GB / core is available) or unnecessarily restrictive (if more than 2 GB / core is available)

cupid/run.py
```diff
@@ -83,110 +86,75 @@ def run(
...
-                                ),
-                            ])
-                    else:
-                        ts_output_dirs = [
-                            os.path.join(
-                                    timeseries_params["ts_output_dir"],
-                                    f"{component}", "proc", "tseries",
-                            ),
-                        ]
+                    year_start = int(year_start[0])
```
mnlevy1981, line 131: The current implementation is that `start_years` (and `end_years`) should be lists that are the same length as [`case_name`](https://github.com/NCAR/CUPiD/blob/main/examples/key_metrics/config.yml#L60). This lets us, as an example, compare 60 years of a current run against 100 years of a baseline.

cupid/run.py
```diff
@@ -83,110 +86,75 @@ def run(
...
-                    num_procs,
-                    serial,
-                    logger,
-                )
-                # fmt: on
-                # pylint: enable=line-too-long
+                    year_end = int(year_end[0])
+
+            modb = gents.ModelOutputDatabase(
+                hf_head_dir=global_params["CESM_output_dir"] + "/" + timeseries_params["case_name"],
```
mnlevy1981, line 142: What happens if `timeseries_params["case_name"]` is a list?  As mentioned above, this is the case in our `key_metrics` example, and we want to loop through each `case_name`, applying the appropriate `start_year` and `end_year` (`ts_done` and `overwrite_ts` are both lists of the same length as well, though I'm not 100% clear on why `ts_done` is in the config file)

mnlevy1981 (changes requested):
I haven't had a chance to test this out yet, but a few things jumped out at first glance. I think this will be a great step forward for us, but we want to make sure we don't lose any of the existing functionality :)

cupid/run.py
```diff
@@ -83,110 +86,75 @@ def run(
...
-                # -----
-
-                # fmt: off
-                # pylint: disable=line-too-long
-                cupid.timeseries.create_time_series(
-                    component,
-                    timeseries_params[component]["vars"],
-                    timeseries_params[component]["derive_vars"],
-                    timeseries_params["case_name"],
-                    timeseries_params[component]["hist_str"],
```
DominikStiller, line 171: It looks like with GenTS there is no way to select the history files to process. For example, a user may just want to create timeseries for the primary history (h0) but not auxiliary files (h1, h2, ...).

AgentOxygen: I am currently developing a new API for GenTS that should work better with CUPiD and will be sure to include this as a feature

Issue #146, opened 2024-10-16
Move contributor's guide to documentation


TeaganKing:
We should also add notes on how to hide cells in the 'how to add notebooks' page.
Link to contributor's guide in 'how to add notebooks' page.
While doing this, we should also include suggestions for notebook names (see #96)

Issue #96, opened 2024-04-25
Standardize notebook names?


TeaganKing:
It may be useful to have some standard notebook names (eg, obs comparisons, basic diagnostics, etc) across components.

One issue that prompted this is that seaice has a notebook titled seaice, which is confusing in the config file.

TeaganKing:
A suggestion could be added to the contributing wiki page & existing notebook names could be updated

TeaganKing:
One potential naming convention could be as follows: `<region>_<variable>_<metric>_<comparisons>.ipynb`

Notebooks will already be categorized under a header for, eg, `atm`, so the component is not needed in the title of the notebook.

Some examples of this might be renaming `NMSE_PSL.ipynb` and `LIWG_SMB_diagnostic.ipynb` to `Global_PSL_NMSE_compare_obs_lens.ipynb` and `Greenland_SMB_visual_compare_obs.ipynb`, respectively.

Issue #60, opened 2024-01-26
Jupyter Books without code.


dabail10:
It would be really nice to have an option to hide the code on the final Jupyter Book HTML. This is nicer for the scientists to look at.

kafitzgerald:
This should be fairly easy to do: 
* https://jupyterbook.org/en/stable/interactive/hiding.html
* https://jupyterbook.org/en/stable/content/metadata.html#jupyter-cell-tags

I'm guessing the hidden option would be better than fulling removing the full code cells.

PR #147, opened 2024-10-17
Documentation updates: contributor's guide, notebook name standardization, hiding code cells

TeaganKing:
This PR will address issues raised in #146, #96, and #60. The key items it addresses are as follows:
1. Provide documentation on how to hide code cells
2. Provide suggestions for notebook names
3. Move contributor's guide from wiki to documentation page
4. Link to contributor's guide in 'how to add notebooks' page
5. Update a few outdated links and clarify a few steps
6. Clarify instructions on pre-commit installation
7. Update PR checklist to make pre-commit clearer and remove irrelevant items

Still need to do:
- [x] Remove wiki contributor's guide once this is merged in
- [x] Check with CUPiD team that notebook naming suggestion is reasonable (will check in next meeting and can edit later, too)
- [x] Hide any additional cells in `key_metrics` examples that need hiding

* [x] Does your submission pass tests?
* [x] Have you lint your code locally prior to submission?


TeaganKing:
Here is a screenshot of the build that implements the current changes.

Screen Shot 2024-10-17 at 9 29 41 AM

docs/conf.py
```diff
@@ -18,12 +18,12 @@
...
 print("sys.path:", sys.path)
 
-# Copy README, NCAR_tips, and contributors into docs
+# Copy README, NCAR_tips, and contributors, and Contributors_Guide into docs
 # This block allows us to remove the header image from any md files
 # without affecting the original version, but still pull the source
 # into the docs build fresh each time.
 
-for file in ["README.md", "NCAR_tips.md", "contributors.md"]:
+for file in ["README.md", "NCAR_tips.md", "contributors.md", "Contributors_Guide.md"]:
```
mnlevy1981, line 26: Three things:

1. `Contributors_Guide.md` is in `docs/`, not `${CUPID_ROOT}/`, so I don't think it needs to be copied here.
2. Should we rename `contributors.md` -> `contributors_to_cupid.md` to match the name change in the rst file? (If so, the rst file needs to be updated to point to the new name of the markdown file). Similarly, do we want to rename `ContributorsGuide.rst` -> `Contributors_Guide.rst` to match the markdown file?
3. Getting off on a tangent here, but we should also stick to a convention about underscores and capitalization. Maybe it should be `contributors_guide.rst` and `contributors_guide.md`? Then `adding_notebooks_to_collection.md` as well?

docs/addingnotebookstocollection.md
```diff
@@ -40,4 +41,9 @@ Generally, a good fit for a diagnostic notebook is one that reads in CESM output
 
 7. Update your parameters. Parameters that are specific to just this notebook should go under `parameter_groups` in the notebook's entry under `compute_notebooks`. Global parameters that you want passed in to every notebook in the collection should go under `global_params`.  When `CUPiD` executes your notebook, all of these parameters will get put in a new cell below the cell tagged `parameters` that you added in step 3. This means they will supercede the values of the parameters that you put in the cell above---the names, notation, etc. should match to make sure your notebook is able to find the variables it needs.
 
-8. All set! Your collection can now be run and built with `cupid-run` and `cupid-build` as usual.
+8. Your collection can now be run and built with `cupid-run` and `cupid-build` as usual.
```
mnlevy1981, line 44: How about

> Your collection can now be run with `cupid-run`, and then the website can be built with `cupid-build`.

docs/addingnotebookstocollection.md
```diff
@@ -40,4 +41,9 @@ Generally, a good fit for a diagnostic notebook is one that reads in CESM output
...
+
+9. If you're happy with your notebook and want to add it to the CUPiD repository, there are a few formatting items that we would like contributors to follow:
+    * Title your notebook something descriptive. A recommended format is `<region>_<variable>_<metric>_<comparisons>.ipynb`; for instance, this might look like `Global_PSL_NMSE_compare_obs_lens.ipynb` or `Greenland_SMB_visual_compare_obs.ipynb`.
```
mnlevy1981, line 47: Do you want to rename the two key metric notebooks to match this proposed convention? I don't think we have open PRs updating those existing notebooks, and then we can point other open PRs to this one when we suggest new file names

README.md
```diff
@@ -44,11 +44,10 @@ If you do not have `mamba` installed, you can still use `conda`... it will just
    $ pip install -e .  # installs cupid
    ```
 
+1. Run `pre-commit install` to configure `git` to automatically run `pre-commit` checks when you try to commit changes; the commit will only proceed if all checks pass. Note that CUPiD uses `pre-commit` to ensure code formatting guidelines are followed, and pull requests will not be accepted if they fail the `pre-commit`-based Github Action.
```
mnlevy1981, line 47: Should we say something about `pre-commit` being available in the `(cupid-dev)` environment? I think we want to recommend doing all commits from `(cupid-dev)` so `pre-commit` doesn't need to be available in other environments (also, to avoid doing the one-time setup stage of `pre-commit` more than once)

docs/addingnotebookstocollection.md
```diff
@@ -40,4 +41,9 @@ Generally, a good fit for a diagnostic notebook is one that reads in CESM output
...
+    * Add a [cell tag](https://jupyterbook.org/en/stable/content/metadata.html#jupyter-cell-tags) `hide-input` for cells which output plots, and add the tag `hide-cell` for cells that do not contain plots (this will hide both the input and output). Do this through JupyterHub when editing your notebook: click `View --> Cell Toolbar --> Tags` and add either `hide-input` or `hide-cell`. This makes it easier to glance at the plots once the webpage is built and not need to scroll through code cells.
+    * Set up `pre-commit` to ensure that your code is properly formatted and linted. Running `pre-commit install` will configure `git` to automatically run the `pre-commit` checks when you try to commit changes; the commit will only proceed if all the checks pass.
```
mnlevy1981, line 49: similar comment as above, do we need to specify `(cupid-dev)` here? I don't know that it needs to be brought up every time `pre-commit` is written about, but I do want to mention it at least once. I think the README makes more sense than this page, but would be open to mentioning the environment here instead (or in both spots).

mnlevy1981 (changes requested):
This is a huge update for the documentation, and it looks great! A few small tweaks to think about, but overall 👍

README.md
TeaganKing: I specified this and added some more details in the Contributors guide, too.

docs/addingnotebookstocollection.md
TeaganKing: Great idea

docs/conf.py
TeaganKing: 1. Thanks for noting this!
2. I had purposely named these different things (similarly to the existing .md and .rst files) since using the same name here seemed to result in sphinx warnings due to unclear references to duplicate filenames (with different extensions).

TeaganKing: 3. I adjusted the .rst files to use lower case and underscores, and the .md files to use camelcase and no underscores...

TeaganKing:
Thanks for these suggestions, @mnlevy1981 . FYI I also re-built and fixed a few links, and it is building properly. The documentation still looks the same as the above screenshot.

mnlevy1981 (approved):
LGTM!

PR #154, opened 2024-11-21
Adf in key metrics

mnlevy1981:
### Description of changes:

Add `link_to_ADF.ipynb` to the `key_metrics` example. This required a little modification to `generate_adf_config_file.py` to account for the fact that off-line regridding happened in 092 (CESM output on the CAM-SE grid, and ADF needs fields on the FV grid to plot them)

* [x] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/ContributorGuide.html)?


examples/key_metrics/config.yml
```diff
@@ -124,6 +124,18 @@ compute_notebooks:
             regridded_output: False # it looks like output is already on f09 grid, didn't need to regrid time-series file
             base_regridded_output: True
             validation_path: '/glade/campaign/cesm/development/cross-wg/diagnostic_framework/nmse_validation/fv0.9x1.25'
+      link_to_ADF:
+        kernel_name: cupid-infrastructure
+        parameter_groups:
+          none:
+            adf_root: ../../examples/key_metrics/ADF_output/
+            case_year_range: "1_100"
```
TeaganKing, line 132: This formatting is a bit confusing. Maybe we could use [1,100] instead?

TeaganKing:
This may be something on my end, but when I ran this, everything was computed with no issue but an additional (undesirable) `key_metrics` directory appeared here: `CUPiD/examples/key_metrics/computed_notebooks/key_metrics` and thus I ran into some issues generating the webpage. I'll try starting with a clean version since it's possible something is outdated in my sandbox.

mnlevy1981:
Two more questions to consider:
1. Should we add the ADF comparison table to the `Link to ADF Output` notebook?
2. Should we add `TAUY` to the variable list so that we generate a lat-lon vector plot? This should also add a second plot to the output in the notebook

TeaganKing:
Based on our conversation I think it sounded like these were two of the key items that users will want to look at, so I would think yes to both of those questions.

mnlevy1981:
From 02cb0ce:

CUPiD ADF Table

justin-richling:
Looks good @mnlevy1981! I guess the only improvement would be to work on the number formatting. Thanks for doing this

mnlevy1981:
eff58b9 adds a row to the table and the vector map:

CUPiD ADF Table 2

mnlevy1981:
> Based on our conversation I think it sounded like these were two of the key items that users will want to look at, so I would think yes to both of those questions.

Done!



> Looks good @mnlevy1981! I guess the only improvement would be to work on the number formatting. Thanks for doing this

Good catch! That's probably an option in the `pandas` `df.to_html()` function... I'll look closer at the documentation...

mnlevy1981:
I found the ADF call to `to_html()`, and added `float_format='{:6g}'.format` to my argument list. In 5c20647, it looks like:

CUPiD ADF Table 3

TeaganKing (approved):
This looks great!

justin-richling (approved):
This looks good, thanks!

Issue #160, opened 2024-12-12
Don't look if ADF output in cupid-webpage unless cupid-diagnostics ran the atm notebooks


mnlevy1981:
**Describe the bug**
When testing #159, I ran

```
$ cupid-diagnostics --land
$ cupid-webpage
```

and got

```
Traceback (most recent call last):
  File "${CONDA_ENVS}/cupid-infrastructure/bin/cupid-webpage", line 8, in <module>
    sys.exit(build())
             ^^^^^^^
  File "${CONDA_ENVS}/cupid-infrastructure/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "${CONDA_ENVS}/cupid-infrastructure/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "${CONDA_ENVS}/cupid-infrastructure/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "${CONDA_ENVS}/cupid-infrastructure/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "${CUPID_ROOT}/cupid/cupid_webpage.py", line 57, in build
    shutil.copytree(
  File "${CONDA_ENVS}/cupid-infrastructure/lib/python3.11/shutil.py", line 559, in copytree
    with os.scandir(src) as itr:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './ADF_output'
```

**To Reproduce**
Run `cupid-diagnostics` for the `key_metrics` example on a subset of components that does not include the atmosphere, then run `cupid-webpage`

**Expected behavior**
Only try to copy output from the ADF if it exists (or maybe if we expect it to exist?)

dabail10:
Similar issue with cupid-clean:

(cupid-infrastructure) [dbailey@derecho1 ~/CUPiD]> cupid-clean
ERROR: config.yml not found

mnlevy1981:
[...]

PR #165, opened 2025-01-16
Include ILAMB output in notebook and provide link to full diagnostic package

TeaganKing:
### Description of changes:
- Wrote script to generate ILAMB `model_setup.txt` file based on `config.yml`
-  Included `link to ILAMB` notebook which includes a link to ILAMB full output as well as a few 'key metric' plots from ILAMB directly in the jupyterbook
- Included config files for BGC & SP cases (without hardcoded `glade` paths) in a parallel location to ADF config file templates.

* [x] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/ContributorGuide.html)?


TeaganKing:
Note the linked issue also requires workflow updates to include ILAMB and choose BGC or SP.

TeaganKing:
Made some changes per discussion in CUPiD meeting today.
- [x] Added active layer thickness plot
- [x] A few other minor changes (eg, remove word 'link')

Still need to do the following:
- [ ] Add overarching "score chart"
- [ ] Remove `glade` filepaths, specify them in BGC cfg file, and add to helper script to generate these

TeaganKing:
The overarching score chart is not a static image-- rather an html page. Thus, I think it'd be difficult to reproduce as a static image within the CUPiD jupyterbook output. This will also be the first thing that appears when you click on the link to the full ILAMB output, so I'm going to suggest we don't implement this as one of the plots that shows up automatically within the jupyterbook after all.

Also, for single runs, this is not the most interesting score chart...
Screenshot 2025-01-24 at 9 28 19 AM

It's really only useful for comparisons against other runs:
Screenshot 2025-01-24 at 9 30 44 AM

mnlevy1981:
For `key_metrics`, can we run ILAMB on both the new case and the baseline? And is that table useful if there are two columns? If so, I think you can imbed one HTML page into another with [something like this](https://www.w3schools.com/howto/howto_html_include.asp):

```py3
display(
    HTML(
         '<div w3-include-html="/path/to/table.html"></div>'
    )
)
```

But I agree that it's not very useful for the single run case. If the above code works, maybe we could include it in an `if print_table:` block where `print_table` defaults to `False`, just so it is available in the future. If the above code doesn't work, I don't think it's worth the time and agree with

> I'm going to suggest we don't implement this as one of the plots that shows up automatically within the jupyterbook after all

mnlevy1981:
(my first version of the above comment relied on javascript, but I think the HTML `<frames>` option would work better

TeaganKing:
Thanks Mike. Good point. FYI I'm still doing some final testing on making sure this runs smoothly with regridded CLM6 data....

TeaganKing:
Removing `mpiexec` from command works! Updating helper script shortly to implement new script suggestion. For now, we are planning to bring this in to the external diagnostic packages example (and not key metrics until the mpiexec issue and the copying ADF output dir issue are both fixed since the latter will also apply to ILAMB).

TeaganKing:
I'm still getting a VarNotInModel error from ILAMB-- also with the old configurations from the previous ILAMB example. Otherwise this full workflow is working.

TeaganKing:
This worked: `ilamb-run --config ../ilamb_aux/ilamb_nohoff_final_CLM.cfg --build_dir bld/ --df_errs ../ilamb_aux/quantiles_Whittaker_cmip5v6.parquet --define_regions ../ilamb_aux/DATA/regions/LandRegions.nc ../ilamb_aux/DATA/regions/Whittaker.nc --regions global --model_setup ../examples/external_diag_packages/model_setup.txt --filter .clm2.h0.`

When model_setup.txt used this filepath: `/glade/campaign/cgd/tss/common/Land_Only_Simulations/CTSM52_DEV/ctsm51_ctsm51d166deadveg_1deg_CRUJRA_FLDS_ABsnoCDE_blk_A5BCD_hist/lnd/hist/`

and ilamb_nohoff_final_CLM.cfg has filepaths such as `/glade/campaign/cesm/community/lmwg/diag/ILAMB/DATA/biomass/Tropical/biomass_0.5x0.5.nc`

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,128 @@
...
+    base_case_output_dir = os.path.join(
+        c_dict["global_params"]["CESM_output_dir"],
+        c_dict["global_params"]["base_case_name"],
+    )
+    with open(cupid_config_loc + "model_setup.txt", "w") as ms:
+        ms.write(
+            "# Model Name    , Location of Files                                                                    ,  Shift From,  Shift To\n",  # noqa: E501
+        )
+        ms.write(
+            f"CTSM51          , {base_case_output_dir}/lnd/hist/regrid/",
```
TeaganKing, line 101: I should add some sort of "if regrid" here

TeaganKing:
We actually want to remap with this command to avoid steradian units and instead get km^2 for area: `ncremap -m /glade/campaign/cesm/cesmdata/inputdata/cpl/gridmaps/ne30pg3/map_ne30pg3_TO_fv0.9x1.25_blin.240826.nc  -P clm "$file"  regrid/"$file"`

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,128 @@
...
+        )  # TODO: aslo update model name? Add to cupid config file?
+    print(f"wrote {cupid_config_loc}model_setup.txt")
+    print("You can now run ILAMB with the following commands:")
+    print("---")
+    print("qinteractive -l select=1:ncpus=16:mpiprocs=16:mem=100G -l walltime=06:00:00")
+    print("conda activate cupid-analysis")
+    print("export ILAMB_ROOT=../ilamb_aux")
+    if run_type == "SP":
+        print(
+            f"ilamb-run --config ../ilamb_aux/ilamb_nohoff_final_CLM_SP.cfg --build_dir {cupid_config_loc}ILAMB_output/ --df_errs ../ilamb_aux/quantiles_Whittaker_cmip5v6.parquet --define_regions ../ilamb_aux/DATA/regions/LandRegions.nc ../ilamb_aux/DATA/regions/Whittaker.nc --regions global --model_setup {cupid_config_loc}odel_setup.txt --filter .clm2.h0.",  # noqa: E501
```
TeaganKing, line 111: Fix typo "model_setup.txt".

Also include that may need remapping.

helper_scripts/generate_ilamb_config_files.py
TeaganKing: I added a note that regridded files are required

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def generate_ilamb_cfg(cesm_root, cupid_config_loc, run_type):
+    """Create config file for use in ILAMB"""
+    sys.path.append(os.path.join(cesm_root, "cime"))
+
+    cupid_root = os.path.join(cesm_root, "tools", "CUPiD")
+    # Is cupid_config_loc a valid value?
+    if cupid_config_loc is None:
+        cupid_config_loc = os.path.join(cupid_root, "examples", "key_metrics")
```
mnlevy1981, line 50: Should this default be `"external_diag_packages"` instead of `key_metrics` since we won't be able to add `link_to_ILAMB.ipynb` in this PR?

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+    base_case_output_dir = os.path.join(
+        c_dict["global_params"]["CESM_output_dir"],
+        c_dict["global_params"]["base_case_name"],
+    )
+    with open(cupid_config_loc + "model_setup.txt", "w") as ms:
+        ms.write(
+            "# Model Name    , Location of Files                                                                    ,  Shift From,  Shift To\n",  # noqa: E501
+        )
+        ms.write(
+            f"CTSM51          , {base_case_output_dir}/lnd/hist/regrid/\n",
```
mnlevy1981, line 101: Instead of `CTSM51`, do we want this to be `c_dict["global_params"]["base_case_name"]`? Also, I'm surprised we are adding the baseline case but not the experimental case. Should we also have a line with `c_dict["global_params"]["case_name"]`?

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+        cupid_config_loc = os.path.join(cupid_root, "examples", "key_metrics")
+    if not os.path.exists(os.path.join(cupid_config_loc, "config.yml")):
+        raise KeyError(f"Can not find config.yml in {cupid_config_loc}")
+
+    with open(os.path.join(cupid_config_loc, "config.yml")) as c:
+        c_dict = yaml.safe_load(c)
+    base_case_output_dir = os.path.join(
+        c_dict["global_params"]["CESM_output_dir"],
+        c_dict["global_params"]["base_case_name"],
+    )
```
mnlevy1981, line 95: According to the spec, users can specify `c_dict["global_params"]["base_case_output_dir"]` if the baseline is not in the same directory as the experimental case... so I think you want something like:

```py3
if "base_case_output_dir" in c_dict["global_params"]:
    base_case_output_dir = os.path.join(
        c_dict["global_params"]["base_case_output_dir"],
        c_dict["global_params"]["base_case_name"],
    )
else:
    base_case_output_dir = os.path.join(
        c_dict["global_params"]["CESM_output_dir"],
        c_dict["global_params"]["base_case_name"],
    )
```

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+    with open(cupid_config_loc + "model_setup.txt", "w") as ms:
```
mnlevy1981, line 96: this should be `os.path.join(cupid_config_loc, "model_setup.txt")` instead of `cupid_config_loc + "model_setup.txt"` -- `cupid_config_loc` isn't guaranteed to have the trailing `/`

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+        ms.write(
+            "# Model Name    , Location of Files                                                                    ,  Shift From,  Shift To\n",  # noqa: E501
+        )
+        ms.write(
+            f"CTSM51          , {base_case_output_dir}/lnd/hist/regrid/\n",
+        )
+    print(f"wrote {cupid_config_loc}model_setup.txt")
```
mnlevy1981, line 103: Same comment about using `os.path.join` because `cupid_config_loc` might not have the trailing `/`

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,131 @@
...
+    print(
+        f"WARNING: ILAMB requires regridded output to be in {base_case_output_dir}/lnd/hist/regrid/ directory.",
+    )
+    print("You can now run ILAMB with the following commands:")
+    print("---")
+    print("qinteractive -l select=1:ncpus=16:mpiprocs=16:mem=100G -l walltime=06:00:00")
```
mnlevy1981, line 109: The `qinteractive` is only necessary on the NCAR computers. Maybe we could do something like

```py3
print("You can now run ILAMB with the following commands:")
print("(Users on a super computer should make sure they are on a compute node rather than a login node)")
print("---")
```

I'm not sure what to do about the fact that we're having trouble with MPI on derecho and casper; in general, users will need the `mpiexec` but for now it's probably okay to stick with the serial recommendation

examples/external_diag_packages/config.yml
```diff
@@ -192,23 +158,11 @@ book_toc:
 
     - caption: Atmosphere
       chapters:
-        - file: atm/link_to_ADF
-
-    # - caption: Ocean
-    #   chapters:
-    #       - file: ocn/ocean_surface
+       - file: atm/link_to_ADF
```
mnlevy1981, line 161: It looks like the spacing changed here? comparing old line 195 to new line 161 I see

```diff
-         - file: atm/link_to_ADF
+        - file: atm/link_to_ADF
```

examples/key_metrics/config.yml
```diff
@@ -131,6 +131,7 @@ compute_notebooks:
         parameter_groups:
           none:
             adf_root: ../../examples/key_metrics/ADF_output/
+            key_plots: ["Surface_Wind_Stress_ANN_LatLon_Vector_Mean.png", "PRECT_ANN_LatLon_Mean.png", "PS_DJF_SHPolar_Mean.png", "TaylorDiag_ANN_Special_Mean.png"]
```
mnlevy1981, line 134: I'm surprised these weren't already included in the `config.yml`, thanks for getting rid of the hard-coded list and replacing it with something configurable!

nblibrary/lnd/link_to_ILAMB.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "The output from ILAMB is in the link below.\n",
+    "\n",
+    "\n",
+    "Note that ILAMB is currently run by users. We hope to update this in response to [CUPiD issue #105](https://github.com/NCAR/CUPiD/issues/105). Temporarily, we suggest users walk through the following process:\n",
+    "1) Install ILAMB\n",
+    "2) Use the `CUPiD/helper_scripts/generate_ilamb_config_files.py` script to generate ILAMB config files based on a CUPiD configuration file.\n",
+    "   * `cd CUPiD/helper_scripts`\n",
+    "   * `./generate_ilamb_config_file.py --cupid_file ../examples/external_diag_packages/config.yml --out_dir ../../`\n",
+    "4) Run ILAMB with the newly created configuration file.\n",
+    "   * `qinteractive -l select=1:ncpus=16:mpiprocs=16:mem=100G -l walltime=06:00:00`\n",
```
mnlevy1981, line 19: I had recommended removing the `qinteractive` line from the helper script, but for the notebook maybe we can just make a note that this line is specific for NCAR users (and other supercomputers might have different ways to request compute resources)? Also, we can drop to `ncpus=1:mpiprocs=1` while we work out issues in the environment

nblibrary/lnd/link_to_ILAMB.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "   * `conda activate cupid-analysis`\n",
+    "   * `export ILAMB_ROOT=../../ilamb_aux`\n",
+    "   * `mpiexec ilamb-run --config ilamb_nohoff_final_CLM.cfg --build_dir bld/ --df_errs ../../ilamb_aux/quantiles_Whittaker_cmip5v6.parquet --define_regions ../../ilamb_aux/DATA/regions/LandRegions.nc ../../ilamb_aux/DATA/regions/Whittaker.nc --regions global --model_setup model_setup.txt --filter .clm2.h0.`"
```
mnlevy1981, line 22: We should drop the `mpiexec` here (and add it back in once things work)

mnlevy1981 (changes requested):
I didn't try to run this, but read through the diffs and have a few comments. Overall, this looks really good and I'm excited about adding ILAMB to `external_diag_packages`!

Should we address #160 in this PR, since we'll have the same problem with ILAMB and users only running CUPiD on non-land components? Seeing your change to `cupid/cupid_webpage.py` brought that issue to mind, but I'd be happy to wait for a later PR to try to fix it.

mnlevy1981 (changes requested):
I had one comment about cleaning up some of the code in `cupid_webpage.py`, but I also have a general question about the functionality of the new `generate_ilamb_config_files.py`.

I was trying to see how this would fit in with the CESM workflow, so I went into a test case directory and used `generate_cupid_config_for_cesm_case.py` to create `config.yml`. Then I used the new script to generate the config files for ILAMB:

```bash
$ ../../../tools/CUPiD/helper_scripts/generate_ilamb_config_files.py --cesm-root /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid --cupid-config-loc ${PWD} --run-type SP
{'cesm_root': '/glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid', 'cupid_config_loc': '/glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid', 'run_type': 'SP'}
wrote /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/ilamb_nohoff_final_CLM_SP.cfg
wrote /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid/model_setup.txt
WARNING: ILAMB requires regridded output to be in /glade/derecho/scratch/mlevy/archive/b.e23_alpha17f.BLT1850.ne30_t232.092/lnd/hist/regrid/ directory.
You can now run ILAMB with the following commands:
(Users on a super computer should make sure they are on a compute node rather than a login node)
---------
conda activate cupid-analysis
export ILAMB_ROOT=../ilamb_aux
ilamb-run --config /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/ilamb_nohoff_final_CLM_SP.cfg --build_dir /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid/ILAMB_output/ --df_errs /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/quantiles_Whittaker_cmip5v6.parquet --define_regions /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/DATA/regions/LandRegions.nc /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/DATA/regions/Whittaker.nc --regions global --model_setup /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid/model_setup.txt --filter .clm2.h0.
---------
```

The first thing that caught my eye is the location of the output files:

```
wrote /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/tools/CUPiD/ilamb_aux/ilamb_nohoff_final_CLM_SP.cfg
wrote /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid/model_setup.txt
```
`model_setup.txt` showed up in the directory where I ran the generate script, which is good. Shouldn't `ilamb_nohoff_final_CLM_SP.cfg` also be written locally? We may need to also copy the quantiles file locally, then we can change the recommendation from `export ILAMB_ROOT=../ilamb_aux` to set the current working directory to `ILAMB_ROOT`.

```diff
-print("export ILAMB_ROOT=../ilamb_aux")
+print(f"export ILAMB_ROOT={os.getcwd()}")
```

(which was actually the second thing I noticed -- as the script is written, we want `print(f"export ILAMB_ROOT=os.path.join(cesm_root,'tools','CUPiD','ilamb_aux')")`).

I think some of these suggestions for getting better output in the CESM workflow may break things for the CUPiD examples; it'll probably take a little back-and-forth to figure out how we want things to work in both use-cases.

nblibrary/lnd/link_to_ILAMB.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "# Link to ILAMB output\n",
+    "\n",
+    "The output from ILAMB is in the link below.\n",
+    "\n",
+    "\n",
+    "Note that ILAMB is currently run by users. We hope to update this in response to [CUPiD issue #105](https://github.com/NCAR/CUPiD/issues/105). Temporarily, we suggest users walk through the following process:\n",
+    "1) Install ILAMB\n",
+    "2) Use the `CUPiD/helper_scripts/generate_ilamb_config_files.py` script to generate ILAMB config files based on a CUPiD configuration file.\n",
+    "   * `cd CUPiD/helper_scripts`\n",
+    "   * `./generate_ilamb_config_file.py --cupid_file ../examples/external_diag_packages/config.yml --out_dir ../../`\n",
```
mnlevy1981, line 17: 1. The file has been renamed from `generate_ilamb_config_file.py` to `generate_ilamb_config_files.py` (files, plural)
2. Command line arguments have changed:

   ```bash
   (cupid-infrastructure) $ ./generate_ilamb_config_files.py --cupid_file ../examples/external_diag_packages/. config.yml --out_dir ../../
   usage: generate_ilamb_config_files.py [-h] --cesm-root CESM_ROOT [--cupid-config-loc CUPID_CONFIG_LOC]
                                         --run-type RUN_TYPE
   generate_ilamb_config_files.py: error: the following arguments are required: --cesm-root, --run-type
   ```

nblibrary/lnd/link_to_ILAMB.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "4) Run ILAMB with the newly created configuration file.\n",
+    "   * `qinteractive -l select=1:ncpus=1:mpiprocs=1:mem=100G -l walltime=06:00:00` (specific for NCAR users)\n",
+    "   * `conda activate cupid-analysis`\n",
+    "   * `export ILAMB_ROOT=../../ilamb_aux`\n",
+    "   * `ilamb-run --config ilamb_nohoff_final_CLM.cfg --build_dir bld/ --df_errs ../../ilamb_aux/quantiles_Whittaker_cmip5v6.parquet --define_regions ../../ilamb_aux/DATA/regions/LandRegions.nc ../../ilamb_aux/DATA/regions/Whittaker.nc --regions global --model_setup model_setup.txt --filter .clm2.h0.`"
```
mnlevy1981, line 22: I think we now want `export ILAMB_ROOT=../ilamb_aux` (remove one set of `../`)? Also, for the second line we should make use of `$ILAMB_ROOT`:

`ilamb-run --config ilamb_nohoff_final_CLM.cfg --build_dir bld/ --df_errs ${ILAMB_ROOT}/quantiles_Whittaker_cmip5v6.parquet --define_regions ${ILAMB_ROOT}/DATA/regions/LandRegions.nc ${ILAMB_ROOT}/DATA/regions/Whittaker.nc --regions global --model_setup model_setup.txt --filter .clm2.h0.`

Also, it looks like `generate_ilamb_config_files.py` now includes either `_SP` or `_BGC` in the `.cfg` file name, so it should be `--config ilamb_nohoff_final_CLM_SP.cfg`

helper_scripts/generate_ilamb_config_files.py
```diff
@@ -0,0 +1,162 @@
...
+
+    with open(os.path.join(cupid_config_loc, "config.yml")) as c:
+        c_dict = yaml.safe_load(c)
+    case_output_dir = os.path.join(
+        c_dict["global_params"]["CESM_output_dir"],
+        c_dict["global_params"]["case_name"],
+    )
+    if "base_case_output_dir" in c_dict["global_params"]:
+        base_case_output_dir = os.path.join(
+            c_dict["global_params"]["CESM_output_dir"],
```
mnlevy1981, line 115: This should be `c_dict["global_params"]["base_case_output_dir"]`

mnlevy1981 (changes requested):
I'm still having trouble running this in a CESM case, though it's getting closer. I'm playing with this in `/glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid/cases/test_cupid_30b04.005/cupid` and, with the changes mentioned in the individual comments

```bash
${CESM_ROOT}/tools/CUPiD/helper_scripts/generate_ilamb_config_files.py --cesm-root /glade/work/mlevy/codes/CESM/cesm3_0_beta04_cupid --cupid-config-loc . --run-type SP
```

works as expected, but

```bash
export ILAMB_ROOT=ilamb_aux
ilamb-run --config ilamb_nohoff_final_CLM_SP.cfg --build_dir bld/ --df_errs ${ILAMB_ROOT}/quantiles_Whittaker_cmip5v6.parquet --define_regions ${ILAMB_ROOT}/DATA/regions/LandRegions.nc ${ILAMB_ROOT}/DATA/regions/Whittaker.nc --regions global --model_setup model_setup.txt --filter .clm2.h0.
```

Results in lots of `VarNotInModel` messages.

The good news is that I tried to generate a CUPiD config file based on `external_diag_packages/config.yml`, and realized one of the issues I was having was that `b.e23_alpha17f.BLT1850.ne30_t232.092` doesn't have regridded output while `f.e23_alpha17f.FLTHIST_ne30.roughtopo.099` does. So I can sort of run this with CESM output, except my new run doesn't have regridded output either :)

Can we try to chat on Monday to iron out these last issues before merging in?

mnlevy1981 (approved):
LGTM!

PR #185, opened 2025-02-04
Further plot updates and fix observed contour lines

dabail10:
### Description of changes:
* [X] Please add an explanation of what your changes do and why you'd like us to include them.

- Fix the observed contour lines in the spatial plots.
- add back in import of nc_time_axis
- Add the "maximum" ice concentration for contour plots and Labrador Sea plots

#### All PRs Checklist:
* [X] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/ContributorGuide.html)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [X] Have you made sure that the [`pre-commit` checks passed (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html)?
* [X] Have you successfully tested your changes locally?

* [X] Have you [hidden the code cells (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html) in your notebook?
* [X] Have you removed any unused parameters from your cell tagged with `parameters`? These can cause confusing warnings that show up as `DAG build with warnings`.
* [X] Have you moved any observational data that you are using to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data` and ensured that it follows this format within that directory: `COMPONENT/analysis_datasets/RESOLUTION/PROCESSED_FIELD_TYPE`?


nblibrary/ice/plot_diff.py
```diff
@@ -84,9 +21,7 @@ def plot_diff(field1, field2, levels, case1, case2, title, proj, TLAT, TLON):
 
     path_nsidc = "/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data/ice/"
 
-    ds_obs = xr.open_dataset(path_nsidc + "SSMI.ifrac.1981-2005monthlymean.gx1v5.nc")
-
-    ds_pop = add_cyclic(ds_obs)
+    ds_obs = xr.open_dataset(path_nsidc + "sst_HadOIBl_bc_1x1_climo_1980_2019.nc")
```
mnlevy1981, line 24: It looks like you added the 1x1 files to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data/ice`; can we move them to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data/ice/analysis_datasets/1x1`? (The `analysis_datasets/` directory exists, but the `1x1/` subdirectory does not)

mnlevy1981 (changes requested):
Besides the comment about the directory structure of the obs datasets, I'm a little confused about how `path_nsidc` is defined inside `plot_diff.py`. Can this be an argument passed in from the notebook, so this function works on computers that don't have access to campaign store? A couple of related questions:

1. is `path_nsidc` the right variable name? (Is this still NSIDC data? Isn't `HadOIBl` coming from UKMO?)
2. Should we have something like `obs_data_root` that is set to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data/ice/analysis_datasets` in `config.yml` and then we build up all the other `path_` variables from that? `path_nsidc = os.path.join(obs_data_root, 'hemispheric_data', 'NSIDC_SeaIce_extent')`, `path_hadoibl = os.path.join(obs_data_root, '1x1', 'HadOIBl')`, etc?

This might require a little clean up in the other ice notebooks and `.py` files, but it will set a good precedent for later updates

dabail10:
Ok. I have merged Alice's changes and I think this is ready for review.

TeaganKing:
Hi @dabail10 , thanks for making the parameter change. I actually already did that in #191 , and we'll still need the first part of the parameters (eg, parameter = "" #comment instead of #parameter=""). Do you mind removing the change you just made?

dabail10:
Whoops. I missed that. I think we might have some conflicts. Can you remove these from your PR to keep it separated?

TeaganKing:
We'll still need the first part of the parameters (eg, parameter = "" #comment instead of #parameter="")-- thus, I think your changes listed here will not actually work with the parameters cell. Also, this is really more related to the observational repository changes than to this PR, so I'd prefer to leave it in #191. I'll remove the changes here though.

mnlevy1981:
I think merging the latest `main` (including #191) into this branch should raise a conflict that we can resolve by keeping the version that currently appears on `main`

dabail10:
I think this is ready. I have fixed the parameter boxes and now use obs_data_dir.

mnlevy1981 (changes requested):
I focused on the `Hemis_seaice_visual_compare_obs_lens.ipynb` notebook, because that's what gets run in `key_metrics`, and didn't look closely at `Hemis_seaice_visual_compare_contour.ipynb`. I have a few minor comments about the `obs_lens` notebook:

1. `path_model` is in `config.yml` but not in the default parameter cell
2. `obs_data_dir` is used in the definition of `path_nsidc` but also appears in calls like `os.path.join(obs_data_dir, path_nsidc, "N_01_extent_v3.0.csv")`; I think you just want `path_nsidc = os.path.join("ice", "analysis_datasets", "hemispheric_data", "NSIDC_SeaIce_extent")` (formatted however `pre-commit` decides it should be formatted)

Also, I'll fix this in a separate PR but currently these notebooks don't support `CESM_output_dir` and `base_case_output_dir` being different

mnlevy1981 (approved):
Thanks for making these changes, this looks like it is ready to merge!

PR #191, opened 2025-02-18
Include global level obs repo directory

TeaganKing:
### Description of changes:
It would be beneficial to only specify the global obs repo once rather than in every description of particular components' notebook parameters. Then, each component config file section can specify the latter part of the path to their observations.

#### All PRs Checklist:
    * Yes, and this work seems to pass tests, although some uxarray issues are cropping up with Meg's notebook...


nblibrary/ice/Hemis_seaice_visual_compare_obs_lens.ipynb
```diff
@@ -57,20 +57,21 @@
    },
    "outputs": [],
    "source": [
-    "CESM_output_dir = \"/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing\"\n",
```
TeaganKing, line 60: Hi @dabail10 , would it be alright with you if we remove these hard coded parameters from the cell, and instead make it clear that they come from the config file by just providing empty strings? @mnlevy1981 said that you were perhaps wanting to run this notebook outside of CUPiD as well, but I'm thinking you probably won't need these particular defaults anyways, and they may add more confusion.

TeaganKing: Dave and I discussed, and I'll comment these out to avoid confusion but leave them as guidance in case someone is running this notebook outisde of CUPiD.

mnlevy1981 (approved):
LGTM!

PR #198, opened 2025-03-05
Update Coupled Model and Delete adf_quick_run.ipynb now that link_to_ADF.ipynb exists

TeaganKing:
This PR removes the old `adf_quick_run` notebook since we now prefer supporting `link_to_ADF.ipynb` and the helper scripts that create the ADF configuration files

#### All PRs Checklist:


mnlevy1981:
`adf_quick_run.ipynb` is still run from `examples/coupled_model/config.yml`:

```yaml
    atm:
      adf_quick_run:
        parameter_groups:
          none:
            adf_path: ../../../externals/ADF
            config_path: .
            config_fil_str: "config_f.cam6_3_119.FLTHIST_ne30.r328_gamma0.33_soae.001.yaml"
```

We've talked about encouraging folks to use `key_metrics` instead of `coupled_model`, but until we get rid of the latter (which should be done, since it doesn't follow the same variable name conventions as `key_metrics` and that is confusing) I think we should keep the ADF quick run notebook.

Another concern is that maybe @justin-richling, @brianpm, or @nusbaume want to provide an example of invoking ADF from a notebook instead of the command line. If that's not the case, maybe we could remove the quick run notebook but run a different atmosphere notebook in `coupled_model`?

mnlevy1981 (changes requested):
Should've marked "request changes" with https://github.com/NCAR/CUPiD/pull/198#issuecomment-2702269587

nusbaume:
Hi @mnlevy1981 @TeaganKing, after discussing it in an ADF meeting we decided that there isn't really a need to have an example of the ADF being called from a notebook (calling it directly in python or from the command line is fine with us), so feel free to move ahead with this particular PR.  Thanks for checking!

TeaganKing:
Thanks @nusbaume for this feedback. @mnlevy1981 I removed this from the coupled model config file, too. That said, I'm not in any rush to get this in, and it could certainly be done as a larger removal (or updating) of the coupled model example.

TeaganKing:
The past few commits address these changes:
- The `register_cmap` error is coming from [`mom6-tools`](https://github.com/NCAR/mom6-tools/blob/main/mom6_tools/m6plot.py). There are changes in mom6-tools that already address this deprecation, so this PR also now pulls in a more recent `mom6-tools` external.
- The `robust` argument of `plot()` has also been deprecated and hence removed from the `Global_Land_Compare_Two_Cases` notebook.
- Notebook names have been updated to adhere to our suggested format.
- The 'coupled model' (now 'additional metrics') example configuration file has been updated to be a similar format as the key metrics configuration file.
- Parameter cells have been updated for the ocean and land notebooks such that the proper parameters are used from the configuration file. 

Remaining issues:
- [ ] The land notebook fails on data that does not include 'lon' in line 4: `ds[var].sel(lon=300, lat=-10, method="nearest").plot(hue="case", y="levsoi")`

mnlevy1981:
> The land notebook fails on data that does not include 'lon' in line 4: `ds[var].sel(lon=300, lat=-10, method="nearest").plot(hue="case", y="levsoi")`

Is this a case that was run on the CAM-SE grid instead of the FV grid? I bet something like

```py3
if "lndgrid" in ds[var].dims:
    # find index of lndgrid close to 300 E, 10 S
    ds[var].isel(lndgrid=lndgrid_ind).plot(hue="case", y="levsoi")
else:
    ds[var].sel(lon=300, lat=-10, method="nearest").plot(hue="case", y="levsoi")
```

Where the tricky part is finding the `lndgrid` index closest to (300,-10) [I would do it offline once, and then hard-code the index with a comment about where it came from and how it's only useful for the ne30 grid]; alternatively, we could require this notebook to use regridded output if it was run on the SE grid.

TeaganKing:
I found the closest land indices by finding a minimum non-NaN distance between the landgrid and the lat/lon values.

mnlevy1981 (changes requested):
When I ran `cupid-diagnostics` for this example, I got a netcdf error:

```python3
TypeError: Error: /glade/campaign/cgd/oce/datasets/cesm/tx2_3/mld/deBoyer2004/deBoyer04_MLD_remapped_to_tx2_3.nc is not a valid NetCDF 3 file
            If this is a NetCDF4 file, you may need to install the
            netcdf4 library, e.g.,

            $ pip install netcdf4


ploomber.exceptions.TaskBuildError: Error when executing task 'Global_Ocean_Surface'. Partially executed notebook available at /glade/work/mlevy/codes/CUPiD/examples/additional_metrics/computed_notebooks/ocn/Global_Ocean_Surface.ipynb
ploomber.exceptions.TaskBuildError: Error building task "Global_Ocean_Surface"
```

This is strange, because `netcdf4` is part of the `cupid-analysis` environment... but maybe `scipy.io.netcdf_file()` doesn't support netCDF4 files? Or maybe old versions used to support this file format but new ones don't?

I assume I'm the only one running into this (I just removed and recreated `(cupid-analysis)`, so I'll keep playing with it

mnlevy1981:
Thanks @gustavo-marques! The other comment I have is that `get_BLD()` fails now (I don't have the exact error in front of me) so I might comment that call out of `ocean_surface.ipynb`. When we start talking about adding MOM6 diagnostics to CUPiD, we should revisit that

mnlevy1981:
The `get_BLD()` error was `ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.MaskedArray with masked values`:

```
ValueError                                Traceback (most recent call last)
Cell In[14], line 1
----> 1 get_BLD(ds, "oml", grd, args)

File ${CUPID_ROOT}/externals/mom6-tools/mom6_tools/surface.py:522, in get_BLD(ds, var, grd, args)
    520     fname = 'PNG/BLD/'+str(args.casename)+'_BLD_model_'+str(month)+'.png'
    521   model = np.ma.masked_invalid(bld_model[t,:].values)
--> 522   xyplot(model, grd.geolon, grd.geolat, area=area,
    523          save=fname,
    524          suptitle=ds[var].attrs['long_name'] +' ['+ ds[var].attrs['units']+']', clim=(0,1500),
    525          title=str(args.casename) + ' ' +str(args.start_date) + ' to '+ str(args.end_date))
    527 # JFM, starting from 0
    528 months = [0,1,2]

File ${CUPID_ROOT}/externals/mom6-tools/mom6_tools/m6plot.py:620, in xyplot(field, x, y, area, xlabel, xunits, ylabel, yunits, title, suptitle, clim, colormap, extend, centerlabels, nbins, landcolor, axis, add_cbar, aspect, resolution, sigma, annotate, ignore, save, debug, show, interactive, logscale)
    618   #plt.gcf().subplots_adjust(left=.08, right=.99, wspace=0, bottom=.09, top=.9, hspace=0)
    619   axis = plt.gca()
--> 620 cs = axis.pcolormesh(xCoord, yCoord, maskedField, cmap=cmap, norm=norm)
    621 if interactive: addStatusBar(xCoord, yCoord, maskedField)
    622 if add_cbar: cb = plt.colorbar(cs, ax=axis, fraction=.08, pad=0.02, extend=extend)

File ${CONDA_ENVS}/cupid-analysis/lib/python3.11/site-packages/matplotlib/__init__.py:1521, in _preprocess_data.<locals>.inner(ax, data, *args, **kwargs)
   1518 @functools.wraps(func)
   1519 def inner(ax, *args, data=None, **kwargs):
   1520     if data is None:
-> 1521         return func(
   1522             ax,
   1523             *map(cbook.sanitize_sequence, args),
   1524             **{k: cbook.sanitize_sequence(v) for k, v in kwargs.items()})
   1526     bound = new_sig.bind(ax, *args, **kwargs)
   1527     auto_label = (bound.arguments.get(label_namer)
   1528                   or bound.kwargs.get(label_namer))

File ${CONDA_ENVS}/cupid-analysis/lib/python3.11/site-packages/matplotlib/axes/_axes.py:6522, in Axes.pcolormesh(self, alpha, norm, cmap, vmin, vmax, colorizer, shading, antialiased, *args, **kwargs)
   6519 shading = shading.lower()
   6520 kwargs.setdefault('edgecolors', 'none')
-> 6522 X, Y, C, shading = self._pcolorargs('pcolormesh', *args,
   6523                                     shading=shading, kwargs=kwargs)
   6524 coords = np.stack([X, Y], axis=-1)
   6526 kwargs.setdefault('snap', mpl.rcParams['pcolormesh.snap'])

File ${CONDA_ENVS}/cupid-analysis/lib/python3.11/site-packages/matplotlib/axes/_axes.py:6026, in Axes._pcolorargs(self, funcname, shading, *args, **kwargs)
   6024     if funcname == 'pcolormesh':
   6025         if np.ma.is_masked(X) or np.ma.is_masked(Y):
-> 6026             raise ValueError(
   6027                 'x and y arguments to pcolormesh cannot have '
   6028                 'non-finite values or be of type '
   6029                 'numpy.ma.MaskedArray with masked values')
   6030     nrows, ncols = C.shape[:2]
   6031 else:

ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.MaskedArray with masked values
```

gustavo-marques:
Are you using the most recent [mom6-tools](https://github.com/NCAR/mom6-tools)?

mnlevy1981 (approved):
I made a small change to `Global_Ocean_Surface.ipynb` to comment out a function call in `mom6-tools` that was failing, and I also noticed it currently does not generate any plots within the notebooks. So I think there's some clean-up to do with `mom6-tools` but that shouldn't hold up this PR

TeaganKing:
Great, thanks @mnlevy1981 !

gustavo-marques:
> I traced the issue in [#198 (review)](https://github.com/NCAR/CUPiD/pull/198#pullrequestreview-2747972489) to either `intake` or `xarray`, but the fix is to ask @gustavo-marques to update `/glade/u/home/gmarques/libs/oce-catalogs/reference-datasets.yml`:
> 
> ```diff
>   mld-deboyer-tx2_3v2:
>     description: De Boyer Montegut MLD climatology remaped to tx2_3v2
>     driver: netcdf
>     metadata:
>       grid: tx2_3v2
>       frequency: monthly
>       version: 2008
>     args:
>       urlpath: '/glade/campaign/cgd/oce/datasets/cesm/tx2_3/mld/deBoyer2004/deBoyer04_MLD_remapped_to_tx2_3.nc'
>       chunks: {}
> +     engine: netcdf4
> ```
> 
> for some reason this needs to be explicitly noted, it's no longer figured out while trying to open the file. I'll continue my testing with this fix, but I suspect I'll be approving this PR soon

@mnlevy1981, I had to revert this change because it's breaking my diagnostics. I may need to update intake. Let's talk about this today.

Issue #199, opened 2025-03-07
Bug in Hemis_seaice_visual_compare_obs_lens.ipynb


mnlevy1981:
**Describe the bug**
If `climo_nyears` is larger than the total number of years in the output set (but total number of years in the output set is not 1), the Ice Area and Ice Volume plots don't look right

**To Reproduce**
Run a case for two or three years, but keep `climo_nyears = 35` when you run the sea ice notebooks in `key_metrics`; `test_cupid_30a06b.10` was only run for 3 years so the red line should not span the entire time axis:

![Image](https://github.com/user-attachments/assets/fe7f319b-d053-45cb-9749-762e11dd29f0)

**Expected behavior**
The red line above should connect years 1,2, and 3; the Lab Sea section handles it correctly:

![Image](https://github.com/user-attachments/assets/39eb58fb-af72-4c85-882f-d9a68ea3df4d)

mnlevy1981:
i'm pretty sure the issue is in

```python3
# Set up axes
if first_year > 1:
    model_start_year1 = end_year - len(ds1_mar_nh.time)
    model_end_year1 = end_year
    model_start_year2 = base_end_year - len(ds2_mar_nh.time)
    model_end_year2 = base_end_year
    lens1_start_year = ds_cesm1_aicetot_nh.year[60]
    lens1_end_year = ds_cesm1_aicetot_nh.year[95]
    lens2_start_year = ds_cesm2_aicetot_nh.year[110]
    lens2_end_year = ds_cesm2_aicetot_nh.year[145]
else:
    model_start_year1 = 1
    model_end_year1 = climo_nyears
    model_start_year2 = 1
    model_end_year2 = climo_nyears
    lens1_start_year = 1
    lens1_end_year = 36
    lens2_start_year = 1
    lens2_end_year = 36
```

specifically in the `else` block, we probably want

```python3
model_end_year1 = min(len(ds1_mar_nh.time), climo_nyears)
model_end_year2 = min(len(ds2_mar_nh.time), climo_nyears)
```

Also, the first block may have off-by-one issues that can be fixed by adding one to the two `start_year` variables:

```python3
model_start_year1 = end_year - len(ds1_mar_nh.time) + 1
model_start_year2 = base_end_year - len(ds2_mar_nh.time) + 1
```

dabail10:
I think there needs to be a more general fix here. In several places I do the following. In other words, I should reset climo_nyears earlier to the actual number of years on the files.

```
[...]

PR #202, opened 2025-03-07
Fix climo_nyears and time axis off by one.

dabail10:
### Description of changes:
* [X] Please add an explanation of what your changes do and why you'd like us to include them.

This fixes bug #199. The climo_nyears could be longer than the number of years available. This resets it to the minimum. This applies to both sea ice notebooks. Also added some changes to fix the off-by-one problem on the time axis.

#### All PRs Checklist:
* [X] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/contributors_guide.html)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [X] Have you made sure that the [`pre-commit` checks passed (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html)?
* [X] Have you successfully tested your changes locally?

* [X] Have you [hidden the code cells (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html) in your notebook?
* [X] Have you removed any unused parameters from your cell tagged with `parameters`? These can cause confusing warnings that show up as `DAG build with warnings`.
* [X] Have you moved any observational data that you are using to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data` and ensured that it follows this format within that directory: `COMPONENT/analysis_datasets/RESOLUTION/PROCESSED_FIELD_TYPE`?


nblibrary/ice/Hemis_seaice_visual_compare_obs_lens.ipynb
```diff
@@ -219,6 +219,9 @@
     "ds1_ann = ds1.resample(time=\"YS\").mean(dim=\"time\")\n",
     "ds2_ann = ds2.resample(time=\"YS\").mean(dim=\"time\")\n",
     "\n",
+    "climo_nyears = min(climo_nyears, len(ds1_ann.time))\n",
+    "climo_nyears = min(climo_nyears, len(ds2_ann.time))\n",
```
mnlevy1981, line 223: If you have 100 years of 1 run and 20 years of another, but set `climo_nyears = 35` -- do you expect to compare 35 years of run 1 to the full 20 years of run 2, or do you expect to compare 20 years of both? I think you want the former, because the difference is pretty stark if you only have 3 years of run 2 instead of 20:

seaice three year run 3

dabail10:
Well, I really hope I would never compare less than 20 years. However, your point is valid. I think this is really what I want though. I suppose I could have two climo_nyears for each dataset.

nblibrary/ice/Hemis_seaice_visual_compare_obs_lens.ipynb
dabail10: I always want the same years from the LENS runs and the NSIDC, so I should not use climo_nyears here. The only question would be if I change the climo_nyears for each of the model runs. I think I want to always use the same number of years, because we are using this to average the runs. Hence, I would only use the last climo_nyears in the averages for both.

dabail10:
I do think I have some more time axis stuff to come, so if you are happy with this, I would like it merged for now.

mnlevy1981 (approved):

PR #210, opened 2025-03-31
Include ILAMB in key_metrics and CESM workflow

TeaganKing:
This PR includes running ILAMB in key_metrics and as part of the CESM workflow. The default will be having this feature turned off.

### Description of changes:

#### All PRs Checklist:
* [x] Have you tested your changes as part of CESM workflow?


TeaganKing:
Update on testing in `cesm_alpha06d`: The ILAMB html pages are generating and the correct steps are being taken in case.cupid to run ILAMB; however, there seem to be a number of errors (eg, see below) that result in the actual plots not being included in any of the original ILAMB html pages.

```
[DEBUG][0][WorkConfront][SurfaceUpwardLWRadiation/FLUXNET2015][b.e23_alpha17f.BLT1850.ne30_t232.092]
  File "/glade/work/tking/conda-envs/cupid-analysis/bin/ilamb-run", line 563, in WorkConfront
    c.confront(m)
  File "/glade/work/tking/conda-envs/cupid-analysis/lib/python3.11/site-packages/ILAMB/Confrontation.py", line 433, in confront
    obs, mod = self.stageData(m)
               ^^^^^^^^^^^^^^^^^
  File "/glade/work/tking/conda-envs/cupid-analysis/lib/python3.11/site-packages/ILAMB/Confrontation.py", line 341, in stageData
    obs = Variable(
          ^^^^^^^^^
  File "/glade/work/tking/conda-envs/cupid-analysis/lib/python3.11/site-packages/ILAMB/Variable.py", line 277, in __init__
    self.depth_bnds = _createBnds(self.depth)
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "/glade/work/tking/conda-envs/cupid-analysis/lib/python3.11/site-packages/ILAMB/Variable.py", line 37, in _createBnds
    x_bnds[+1:, 0] = 0.5 * (x[:-1] + x[+1:])
                     ~~~~^~~~~~~~~~~~~~~~~~~
TypeError: can't multiply sequence by non-int of type 'float'
```

I'm planning to dig into the files resulting from my case and the 092 case to hopefully find some insights into why this is happening.

TeaganKing:
When I use a different `model_setup.txt` file to specify runs that have been regridded and are known to work with ILAMB (`b.e30_beta02.BLT1850.ne30_t232.104` and `f.e23_alpha17f.FLTHIST_ne30.roughtopo.099`), ILAMB runs smoothly through the `case.cupid` cesm workflow.

However, the link_to_ILAMB.ipynb still shows a path for ilamb_root that includes `external_diag_packages`; this should be updated.
Screenshot 2025-04-15 at 6 14 44 PM
Hence, the plots are not displaying inline.

I would like to also note that there are a handful of additional notebook errors that I avoided by commenting out sections of the config file. I don't think these are relevant to this PR, but I recorded the error output here: `/glade/work/tking/cupid_project/other_cupids/alpha6d/cime/scripts/ilamb_test_newcheckout/cupid-log-copy-from-terminal`

TeaganKing:
This does not work out of the box as is seen here with the examples, but it DOES work with the added `, 50, 2000` on both lines of `model_setup.txt`. So, perhaps we should include an 'if BLT1850 is in casename &/or basecasename, add `, 50, 2000` on both lines of `model_setup.txt`' to the `generate_ilamb_config_files.py` script.

TeaganKing:
Next step: test running ILAMB in CESM workflow

RunCESM.md
```diff
@@ -13,8 +13,11 @@ Install cupid analysis and infrastructure environments per the [usual setup inst
...
 Adjust any variables that you would like (eg, the base case to compare against, how long to run for, etc) with `./xmlchange <VARIABLE>=<VALUE>`
 
-## Adjust wallclock time
-You can adjust the wallclock time for CUPiD specifically by running the following command: `./xmlchange --subgroup case.cupid JOB_WALLCLOCK_TIME={new time}`. You might want to do this if you are for instance running an example with lots of computationally intensive diagnostic notebooks.
+## Adjust wallclock time if needed
+You can adjust the wallclock time for CUPiD specifically by running the following command: `./xmlchange --subgroup case.cupid JOB_WALLCLOCK_TIME={new time}`. You might want to do this if you are running an example with lots of computationally intensive diagnostic notebooks (for instance, a wallclock time of 6hr may be necessary for ILAMB).
+
+## Adjust memory if needed
+Change the memory required for running CUPiD as needed by running a command like the following `./xmlchange CUPID_MEM_PER_TASK=35`. Note that the default is 10GB per task; ILAMB, for instance, requires this to be updated since the default is not enough memory.
```
mnlevy1981, line 20: To fit the format of the suggestion for changing `JOB_WALLCLOCK_TIME`, this should be `./xmlchange CUPID_MEM_PER_TASK={new memory [in GB]}`. I also wouldn't mention the default, since that might be machine dependent. The second sentence could be something like "For instance, you may want to use 35 GB per task when running ILAMB.

cupid/timeseries.py
```diff
@@ -118,7 +118,7 @@ def create_time_series(
         # End if
 
         # Check if history files actually exist. If not then kill script:
-        if not list(starting_location.glob("*" + hist_str + ".*.nc")):
+        if not list(starting_location.glob("*" + hist_str + "*.nc")):
```
mnlevy1981, line 121: Looking at this, I think we want to add a `.` before `hist_str` rather than removing the `.` after `hist_str`; if `hist_str = "clm2.h0"` then I would not expect to convert both `clm2.h0a` and `clm2.h0i` streams to time series. This will be a little tricky for baseline comparisons given the change in naming convention between CESM2 and CESM3 that recently went into the CESM3 development branch, but the solution is probably to have a mechanism for specifying different streams per case rather than trying to find a common string that matches both the old and new streams in a wildcard search.

cupid/timeseries.py
```diff
@@ -131,7 +131,7 @@ def create_time_series(
         for year in range(start_year, end_year + 1):
             # Add files to main file list:
             for fname in starting_location.glob(
-                f"*{hist_str}.*{str(year).zfill(4)}*.nc",
+                f"*{hist_str}*{str(year).zfill(4)}*.nc",
```
mnlevy1981, line 134: same as above, this should be `f"*.{hist_str}.*{str(year).zfill(4)}*.nc"`

examples/external_diag_packages/config.yml
```diff
@@ -162,6 +161,7 @@ compute_notebooks:
                         'Western US','Eastern US','Europe',
                         'Amazonia','Central Africa','Indonesia',
                         'Brazil','Sahel','Southern Africa','India']
+          defaults_file: ../../externals/LDF/lib/ldf_variable_defaults.yaml
```
mnlevy1981, line 164: Was this overlooked in #226 or does ILAMB need this seemingly-LDF-related variable?

mnlevy1981 (changes requested):
This looks great! A few small recommended changes, and then one question about LDF-specific changes :)

cupid/timeseries.py
TeaganKing: Good point. This makes sense to me to keep the `.` before and after so that the user-defined `hist_str` is used as expected. I'll just link #279 here for reference.

Issue #219, opened 2025-05-14
Time series generation should allow user to specify output directory


mnlevy1981:
**Is your feature request related to a problem? Please describe.**
If a user is trying to compare a run to a baseline in another user's scratch space, then they can not generate time series files because CUPiD wants to write that in a parallel directory.

**Describe the solution you'd like**
Users should be able to specify where the time series files are written, and then notebooks that read time series output should look in this new location instead of basing it on the output root

**Additional context**
@dabail10 does this sum up the issue you ran into at our last hackathon?

dabail10:
This pretty much sums it up.

PR #223, opened 2025-05-29
Addition of Time Series Directory Option

TeaganKing:
### Description of changes:
* These changes will address #219 , allowing users to generate time series files in a directory other than the main CESM_output_dir to allow for comparison across cases that the user has created and cases from another user which the primary user does not have permission to edit.
* This is high priority and will address concerns from @dabail10 and @cecilehannay .
* As we expand to multiple base case options, this PR allows for a situation where person A wants to run CUPiD and compare their case to a case from person B as well as a case from person C; in this case, person A's timeseries will appear in person A's CESM_output_dir, person B and C's cases' ts files will both appear in the specified ts_dir/casename/component/proc/tseries (eg, person A's scratch space).

#### All PRs Checklist:


TeaganKing:
I will test these changes with the following commands:
- [ ] `cupid-timeseries` with `ts_dir` in `config.yml` set to `None`.
- [ ] `cupid-timeseries` with `ts_dir` in `config.yml` set to `/glade/derecho/scratch/tking/archive`.
- [ ] `cupid-diagnostics` with `ts_dir` in `config.yml` set to `None` in order to test all notebooks that have been updated.
- [ ] `cupid-diagnostics` with `ts_dir` in `config.yml` set to `/glade/derecho/scratch/tking/archive` in order to test all notebooks that have been updated.
- [ ] helper_scripts/generate_adf_config_file.py --cesm-root <ROOT> --adf-template <TEMPLATE> --out-file adf_config.yml
- [ ] run as part of CESM workflow with `CUPID_TS_DIR` set to default
- [ ] run as part of CESM workflow with `CUPID_TS_DIR` set to `/glade/derecho/scratch/tking/archive`

TeaganKing:
Irrelevant to this PR, but I also just added a checkbox on the PR template for converting from draft to open to hopefully help people sort out when to use what :) Super brief and non-answer changing, so I figured it was fine to just include in this PR....

cime_config/config_tool.xml
```diff
@@ -95,12 +95,20 @@
 
   <entry id="CUPID_BASELINE_ROOT">
     <type>char</type>
-    <default_value>$DOUT_S_ROOT</default_value>
+    <default_value>$(dirname "$DOUT_S_ROOT")</default_value>
```
mnlevy1981, line 98: Was just using `$DOUT_S_ROOT` causing problems?

helper_scripts/generate_adf_config_file.py
```diff
@@ -48,6 +48,7 @@ def generate_adf_config(cesm_root, cupid_config_loc, adf_template, out_file):
     test_case_name = c_dict["global_params"]["case_name"]
     c_ts = c_dict["timeseries"]
     ts_case_names = c_ts.get("case_name")
+    ts_dir = os.path.join(c_dict["global_params"].get("ts_dir", DOUT))
```
mnlevy1981, line 51: I see from line 46 that this variable already existed (and I probably created it :p ) but I don't like the variable name `DOUT`; could we change that to `CESM_output_dir` to match what's in the config file? It looks like it's only used in a couple of spots now that we have `ts_dir` as well

mnlevy1981 (changes requested):
A couple of very minor changes, but this looks great!

For the `Global_PSL_NMSE_compare` issue with missing lat / lon, I think @MacyCallahan ran into a similar problem and it turned out the time series was still on the CAM-SE grid (it needs to be remapped to the CAM-FV grid because it is compared against a bunch of other cases from on the FV grid). Was this coming from a recent CESM run or one of the cases we've been using for a while? I'm happy to hop on a call and look at the error in more detail this afternoon or tomorrow.

helper_scripts/generate_adf_config_file.py
TeaganKing: Will do!

TeaganKing:
Thanks Mike! This was a newly created case (as of ~2 weeks ago); however, I was still comparing against b.e23_alpha17f.BLT1850.ne30_t232.092. Perhaps I should compare against a more recent case?

mnlevy1981:
> This was a newly created case (as of ~2 weeks ago) using the alpha6d tag

For test cases, I've been including

```
interpolate_nlat = 192
interpolate_nlon = 288
interpolate_output = .true.
```

in `user_nl_cam`; alternatively, you should be able to run

```
$ ncremap -m /glade/campaign/cesm/cesmdata/inputdata/cpl/gridmaps/ne30pg3/map_ne30pg3_TO_fv0.9x1.25_blin.240826.nc ${CASE}.cam.h0a.PSL.000101-000112.nc regrid/${CASE}.cam.h0a.PSL.000101-000112.nc
```

after the fact and then set `regridded_output: true` in  `Global_PSL_NMSE_compare_obs_lens.parameter_groups.none` in your CUPiD config file? (I'm 99% sure that's the right set of arguments for `ncremap` but if it gives an error I can dig up the correct call!)

cime_config/config_tool.xml
mnlevy1981: Oh, right -- the `dirname` command actually does something :)

I just want to make sure putting shell commands in an XML file like that is okay...  @jedwards4b is this a valid default value that CIME will always know how to process, or should we do something else?

TeaganKing:
> > This was a newly created case (as of ~2 weeks ago) using the alpha6d tag
> 
> For test cases, I've been including
> 
> ```
> interpolate_nlat = 192
> interpolate_nlon = 288
> interpolate_output = .true.
> ```
> 
> in `user_nl_cam`; alternatively, you should be able to run
> 
> ```
> $ ncremap -m /glade/campaign/cesm/cesmdata/inputdata/cpl/gridmaps/ne30pg3/map_ne30pg3_TO_fv0.9x1.25_blin.240826.nc ${CASE}.cam.h0a.PSL.000101-000112.nc regrid/${CASE}.cam.h0a.PSL.000101-000112.nc
> ```
> 
> after the fact and then set `regridded_output: true` in `Global_PSL_NMSE_compare_obs_lens.parameter_groups.none` in your CUPiD config file? (I'm 99% sure that's the right set of arguments for `ncremap` but if it gives an error I can dig up the correct call!)

I'll test this out when machines are back...

jedwards4b:
aren't you getting the parent directory to add back the $CASENAME?  Can you check if the last part of the directory is the $CASENAME and only add it if it isn't already there?

TeaganKing:
We're sometimes adding the base case name, and sometimes adding the casename. I can try to change this  with my proposed solution above, though.

TeaganKing:
Currently testing using "/.." ...

nblibrary/atm/Global_PSL_NMSE_compare_obs_lens.ipynb
```diff
@@ -165,6 +165,35 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "caf05d8b-a711-40fb-b88a-c10472a49d30",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "if regridded_output:\n",
+    "    file_path = f\"{ts_dir}/{case_name}/atm/proc/tseries/regrid\"\n",
+    "else:\n",
+    "    file_path = f\"{ts_dir}/{case_name}/atm/proc/tseries\"  # not sure why these don't have lat/lon?\n",
```
mnlevy1981, line 178: Can you remove the comment from this line? :)

mnlevy1981:
> I did test it in all the ways described above but I think that given that this is a fairly major update, it would probably be good to make sure it runs for someone else, too?

I'll set up a CESM sandbox with these changes and run a small test. I suspect that run will take a few hours, so I might not approve / merge this PR until tomorrow

cime_config/config_tool.xml
```diff
@@ -95,12 +95,20 @@
 
   <entry id="CUPID_BASELINE_ROOT">
     <type>char</type>
-    <default_value>$DOUT_S_ROOT</default_value>
+    <default_value>$DOUT_S_ROOT/..</default_value>
```
mnlevy1981, line 98: I ran a case on Thursday, and then didn't get a chance to look at the output until just now. My `case.cupid` failed because I needed to change this variable (`./xmlchange CUPID_BASELINE_ROOT=/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing`); I'm re-running `case.cupid` with this change, and if everything else runs successfully then we should talk about what we want this default to be... I'm not sure what the most common use case will be, I suspect it will be comparing against some control runs saved on campaign but probably not in our `cross-wg/diagnostic_framework` directory. So maybe the existing default is okay, but we'll want to change it later on?

mnlevy1981 (approved):
I ran into a couple of problems running the CESM workflow with this branch, but I think we already have plans to fix them elsewhere:

1. I did not configure CAM to output on the fv0.9x1.25 grid, so the Global PSL notebook threw an `AttributeError: 'DataArray' object has no attribute 'lon'`; #250 and #251 should address this
2. I ran out of wallclock time; we already document [how to request more time](https://github.com/NCAR/CUPiD/blob/399fc10fceb61fe9f3dc7ecebc16116d025583d1/RunCESM.md?plain=1#L17), and also we plan to make it easier to run notebooks in parallel with #256 plus related CIME / ccs_config PRs

Given all that, I'm happy to merge this in and then address these concerns in the appropriate PRs

PR #251, opened 2025-07-15
Add remapping to CESM workflow

### LINKED ISSUE #211: Create a cupid-remap script for regridding
**Is your feature request related to a problem? Please describe.**
Similar to how `cupid-timeseries` just calls `ncrcat`, `cupid-remap` could call `ncremap`. The config file should include the path of the mapping file to use for each component (and the CESM workflow will need relative paths in `env_postprocessing.xml` to append to `DIN_LOC_ROOT`

**Describe the solution you'd like**
The script needs to account for the following:

1. CLM grids will call `ncremap -P clm`
2. We will need to update the CESM workflow to add this step (and make sure ATM notebooks know whether to look in `atm/proc/tseries` or `atm/proc/tseries/regrid`)
3. Update documentation

**Additional context**
At some point we will want a python-based regridding script, so the python file we create in `cupid` might change but the workflow updates will still be useful

MacyCallahan:
### Description of changes:
Add remapping to CESM workflow

#### All PRs Checklist:


examples/key_metrics/config.yml
```diff
@@ -154,7 +153,7 @@ compute_notebooks:
     rof:
       global_discharge_gauge_compare_obs:
         parameter_groups:
-          none:
+           none:
```
mnlevy1981, line 156: This whitespace change is unnecessary

nblibrary/atm/Global_PSL_NMSE_compare_obs_lens.ipynb
```diff
@@ -214,8 +214,16 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "dfe0c214",
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
mnlevy1981, line 226: Did one cell get split into two cells in this notebook? It doesn't seem necessary for this PR

mnlevy1981 (changes requested):
A couple small things I noticed that I'll fix before I start testing this (I'm going to wait for `cesm3_0_alpha07c` and then create a multi-year case to run tests in)

TeaganKing:
I'm currently testing this with cesm3_0_alpha07c and cime6.1.118:
- [ ] 1) with remapping on: 
- [ ] 2) with remapping off:

TeaganKing:
`reorder-python-imports` seems to undo the exact changes that `black` makes, and I think `black` has much more general impact on code formatting, so I'd personally prefer to use that one. Let me know if there are alternative suggestions!

helper_scripts/generate_cupid_config_for_cesm_case.py
```diff
@@ -182,6 +190,7 @@ def generate_cupid_config(
     my_dict["global_params"]["base_start_date"] = cupid_base_startdate
     my_dict["global_params"]["base_end_date"] = cupid_base_enddate
     my_dict["timeseries"]["case_name"] = [case, cupid_baseline_case]
+    my_dict["global_params"]["regridded_output"] = cupid_remap
```
mnlevy1981: This should go into `my_dict["global_params"]["compute_notebooks"]["atm"]["Global_PSL_NMSE_compare_obs_lens"]["parameter_groups"]["none"]["regridded_output"]` and maybe also `my_dict["global_params"]["compute_notebooks"]["atm"]["ADF"]["external_tool"]["regridded_output"]` (if they exist) rather than the global parameter

nblibrary/ice/Hemis_seaice_visual_compare_contour.ipynb
```diff
@@ -156,24 +158,43 @@
...
+    "            ts_dir,\n",
+    "            base_case_name,\n",
+    "            \"ice\",\n",
+    "            \"proc\",\n",
+    "            \"tseries\",\n",
+    "            \"regrid\",\n",
+    "            f\"{base_case_name}.cice.h.*.nc\",\n",
+    "        )\n",
+    "    else:\n",
+    "        base_file_path = f\"{ts_dir}/{base_case_name}/atm/proc/tseries\"\n",
```
mnlevy1981: (a) have we tested out this ice notebook with regridded output?
(b) if we really want this block of code, this should be `/ice/proc` rather than `atm/proc`.

helper_scripts/generate_cupid_config_for_cesm_case.py
mnlevy1981: alternatively, maybe we should keep this in global parameters and just remove the local flags from the NMSE and ADF parameters? The ADF will be tricky, because it's not coming through parameter group... but further along in the review I noticed there's some regrid code in the ice notebook as well

mnlevy1981 (changes requested):
I tried running this through the CESM workflow and it doesn't seem like the remapping step actually ran, even though the log says

```
INFO:util:
  Regridding atm time series files...
INFO:util:       Regridding time series for case 'test_cupid_30a07f.020' :
INFO:util:       Regridding time series for case 'b.e30_alpha07c_cesm.B1850C_LTso.ne30_t232_wgx3.228' :
INFO:util:  ... atm time series file remapping has finished successfully.
```

Oh, it's strange that we switch from regridding to remapping in the last line; we should definitely be consistent! My vote would be to only use remap / remapping and replace all instances of regrid / regridding

mnlevy1981:
I grabbed a decent amount of code from this PR and merged it in via #409 -- sorry to take so long to get your contributions in, @MacyCallahan!

PR #302, opened 2025-10-16
enable different time ranges for ts and climo files in LDF

wwieder:
### Description of changes:
LDF allows users to define seperate time bounds for climo vs. ts generation

#### All PRs Checklist:
@TeaganKing volunteered to functionally test this for me
Note, LDF will fail if there are multiple climo files for a variable in the same directory

It may be nice to use the same start year and end year for both ts and climo if users don't define climo_*_year strings.  LDF already does this, but not sure how to enable it in CUPiD?


TeaganKing:
- [ ] @wwieder is still checking that regional plotting is working as expected

TeaganKing (approved):
Looks great to me! Once Will checks the output of regional plots, this can be merged in.

Issue #301, opened 2025-10-15
Support hist_str being different for case & base case


TeaganKing:
**Is your feature request related to a problem? Please describe.**
Sometimes we will compare a case with a history string such as `h0a` with a case with a history string such as `h0`. This causes an incompatibility when we try to run timeseries or select files with the same history string for both cases.

**Describe the solution you'd like**
`hist_str` and `hist_str_base` should both be specified. Notebooks should be updated to reflect this change. Timeseries generation should also look for the different specified hist strings.

**Additional context**
See #90

Issue #233, opened 2025-06-18
Allow base case name to be None


TeaganKing:
The base case name being set to none does not work, particularly in link_to_ADF.ipynb

Reported by @yanchunhe

TeaganKing:
ADF now can be compared to obs with no basename specified; however, we will need to check this functionality for other notebooks.

PR #305, opened 2025-10-17
Allow for different hist strings for base case & test case

TeaganKing:
### Description of changes:
This addresses #301 . It would probably also be useful to address #233 in this PR...
Also, perhaps we should rename `case` and variables with that prefix to `test_case` since this seems to be a confusing point for users.

#### All PRs Checklist:
* [ ] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/contributors_guide.html)?
* [ ] Have you successfully tested your changes locally when running standalone CUPiD?
* [ ] Once you are ready to have your PR reviewed, have you changed it from a Draft PR to an Open PR?

### DIFF
--- a/cupid/run_remapping.py
+++ b/cupid/run_remapping.py
@@ -166,7 +166,8 @@ def run_remapping(
                 component,
                 timeseries_params[component]["vars"],
                 timeseries_params["case_name"],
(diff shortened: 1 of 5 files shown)


TeaganKing:
I think the gist of it was to be able to use h0 for the test case while using h0a for the base case, for instance. Not super important for comparing cases from the same sandbox, but could still be useful to implement.

Issue #316, opened 2025-10-30
Avoid permission hassles


TeaganKing:
Avoid permission hassles in the future (see #272 ) by implementing some sort of change of group & permissions within the tseries/climo generation-- eg something like this:

```
def fix_permissions(path):
    for root, dirs, files in os.walk(path):
        for d in dirs:  # fix dirs
            dpath = os.path.join(root, d)
            os.chmod(dpath, 0o755)  # rwx for owner, rx for group/others
            os.chown(dpath, -1, "cesm")  # change group only

        for f in files:  # fix files
            fpath = os.path.join(root, f)
            os.chmod(fpath, 0o666)  # rw for all
            os.chown(fpath, -1, "cesm")  # change group only
```

PR #321, opened 2025-11-04
TS/Climo updates for permissions

TeaganKing:
This will address #316 .

#### All PRs Checklist:

### DIFF
--- a/cupid/run_timeseries.py
+++ b/cupid/run_timeseries.py
@@ -119,7 +119,7 @@ def run_timeseries(
     file_gid = os.shutil._get_gid(file_group)
     if file_gid is None:
         file_gid = -1
-        f"{file_group} is not a valid group on this machine"
+        print(f"{file_group} is not a valid group on this machine")
         # Or raise an exception because file_group is not defined on this machine
     dir_gid = os.shutil._get_gid(dir_group)
     if dir_gid is None:


TeaganKing:
FYI I didn't change the permissions directly at this step: `list_of_files.append(ts_outfil_str)` because it was still a temporary `.nc.pid*.ncrcat` file. The directory permission updates likely duplicate permission/ownership changes but I'd rather catch everything than not catch everything.

I am wondering if this would pose concerns for users outside of the cesm group. Maybe the GID & permission settings should be configurable with these defaults?

examples/additional_metrics/config.yml
```diff
@@ -58,6 +58,10 @@ timeseries:
   ts_done: [False,False]
   overwrite_ts: [False,False]
   case_name: ['b.e30_beta02.BLT1850.ne30_t232.104', 'b.e23_alpha17f.BLT1850.ne30_t232.092']
+  file_mode: 0o666
+  dir_mode: 0o775
```
mnlevy1981: Do we want 644 for files (only the owner can write) and 775 (owner and anyone else in the group can write, readable for everyone else)?

cupid/run_timeseries.py
```diff
@@ -110,6 +110,10 @@ def run_timeseries(
 
     # general timeseries arguments for all components
     num_procs = timeseries_params["num_procs"]
+    file_mode = timeseries_params["file_mode"]
+    dir_mode = timeseries_params["dir_mode"]
+    file_gid = timeseries_params["file_gid"]
+    dir_gid = timeseries_params["dir_gid"]
```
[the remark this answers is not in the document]
TeaganKing: Hmm. I certainly see the appeal of making sure this is set, but I'm quite wary about setting file permissions without the user being able to easily find where this was set.

cupid/run_timeseries.py
```diff
@@ -28,6 +28,7 @@
 import os
 
 import click
+import shutil
```
mnlevy1981: This should be under `import os` since it's part of the standard library

mnlevy1981 (changes requested):
One more small set of changes, but this is looking good!

PR #340, opened 2025-12-11
Include Notebook for Intermediary CESM Output Diagnostics

TeaganKing:
### Description of changes:
This PR will include a notebook from @cecilehannay that will be a good notebook to look at CESM output during model runs. 
We will still need to do the following:
* [x] Update notebook to use parameters from config file
* [x] pull utils out of notebook
* [x] Either break notebook up into components or create an `intermediary_outputs` notebook directory under nblibrary (I decided breaking into components was the most straightforward, and additionally generalized in order to make it easy to add additional vars)
* [x] in a separate PR, we will want to include capability of running CUPiD intermittently during CESM run (not done yet, but documented in #343)
* [x] make an `intermediary_outputs` example
* [x] This notebook currently expects all tseries to be computed and in one dir (not component-separated). We will need to make sure that this is updated to match the cupid tseries output
* [x] make variables configurable / be able to add to subplots

#### All PRs Checklist:

* [x] Have you removed any unused parameters from your cell tagged with `parameters`? These can cause confusing warnings that show up as `DAG build with warnings`.

### DIFF
--- a/examples/intermediary_outputs/config.yml
+++ b/examples/intermediary_outputs/config.yml
@@ -46,15 +46,6 @@ global_params: # TODO: a lot of these params may be irrelevant/misleading for th
   case_nickname: 'BLT1850_232'
   base_case_nickname: 'BLT1850_228'
   CESM_output_dir: /glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing
-  start_date: '0001-01-01'
-  end_date: '0021-01-01'
-  climo_start_year: 11
-  climo_end_year: 21
-  base_start_date: '0001-01-01'
-  base_end_date: '0045-01-01'
-  base_climo_start_year: 36
-  base_climo_end_year: 45
-  obs_data_dir: '/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data'
   ts_dir: null # If this is set to null, it will default to CESM_output_dir; if you don't have permissions to write to CESM_output_dir, you can specify a directory such as your scratch arcive directory
   lc_kwargs:
     threads_per_worker: 1
--- a/cupid_utils/cupid_utils/multi-component/spatial_means.py
+++ b/cupid_utils/cupid_utils/multi-component/spatial_means.py
@@ -4,39 +4,6 @@ import xarray as xr
 import numpy as np
 
 
-def fix_time_variable(ds):
-    """
-    Description: fix the time of cam file to have the time at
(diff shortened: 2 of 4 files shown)


TeaganKing:
Current output looks like this for atm:
Screenshot 2025-12-12 at 11 23 26 AM
And ICEFRAC is a separate plot.
More vars can be added and number of subplots changes accordingly.

cupid_utils/cupid_utils/generic_fxns.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    time = ds["time"]
+    time = xr.DataArray(
+        ds["time_bnds"].load().mean(dim="nbnd").values,
+        dims=time.dims,
+        attrs=time.attrs,
+    )
+    ds["time"] = time
+    ds.assign_coords(time=time)
+    return ds
```
mnlevy1981: Does CAM still write history files with the time dimension giving the end of the averaging interval? I haven't made it all the way down to the notebooks to see where this is used, but unless we are comparing to CESM1 or CESM2 output I would expect the time dimension to be correct already

cupid_utils/cupid_utils/generic_fxns.py
```diff
@@ -0,0 +1,168 @@
...
+    Compute the weighted annual global mean of
+    a given variable.
+    Argument:
+       case = casename
+       var = variable name
+    """
+    filename = filepath + case + "/" + case + "." + var + ".nc"
+    ds = xr.open_dataset(filename)
+
+    ds.mean(["lat", "time"])
```
mnlevy1981: This mean isn't being stored anywhere, so I don't think it needs to be computed

cupid_utils/cupid_utils/generic_fxns.py
TeaganKing: This isn't used. I can just take it out.

examples/intermediary_outputs/config.yml
```diff
@@ -0,0 +1,178 @@
...
+
+############# NOTEBOOK CONFIG #############
+
+############################
+# Notebooks and Parameters #
+############################
+
+# All parameters under global_params get passed to all the notebooks
+
+global_params: # TODO: a lot of these params may be irrelevant/misleading for this notebook
```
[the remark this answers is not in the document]
TeaganKing: Just updated this. Also realized we should be using ts_dir instead of CESM_output_dir as default for finding ts files, so I updated that, too.

PR #358, opened 2026-01-13
Modify uxarray and pre-commit in cupid envs

TeaganKing:
We want to include updated versions of uxarray and also include pre-commit

- [x] tested building env

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,14 +1,14 @@
 [submodule "externals/ADF"]
-	path = externals/ADF
-	url = https://github.com/NCAR/ADF.git
+    path = externals/ADF
+    url = https://github.com/NCAR/ADF.git
 [submodule "externals/ILAMB"]
-	path = externals/ILAMB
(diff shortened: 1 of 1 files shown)


externals/ILAMB
mnlevy1981: Is this the right commit for `ILAMB`? I thought you needed to back up to something older

mnlevy1981 (changes requested):
Two comments about moving to older versions of packages (ILAMB and uxarray), but if the newer tags have been tested then we should stick with what's in this PR

externals/ILAMB
TeaganKing: Hmm, I thought this would use the 2.7.2 tag: `git submodule add -b v2.7.2 https://github.com/rubisco-sfa/ILAMB.git ./externals/ILAMB`

It looks like it didn't. What is the best way to do this?

TeaganKing: Whoops I clearly didn't reload the page and missed your message before responding again. Thanks for that, though! And yes that is what I ended up doing.

PR #397, opened 2026-03-06
Multi-instance timeseries capability for standalone CUPiD

mollymwieringa:
### Description of changes:

These changes take first steps toward issue #396 by enabling the run_timeseries.py tool to find model history files produced by a multi-instance-enabled case of CESM (e.g. cice_000X.h instead of cice.h). It involves changes to run_timeseries.py and adding a multi-instance example with its own config file to standalone CUPiD. 

These changes will help enable efficient postprocessing of many history files for multi-instances cases into more storage-efficient timeseries files, reducing the amount of overhead required by multi-instance simulations. This benefit is particularly relevant to data assimilation experiments, which use multi-instance simulations to run large ensembles that output history files at high temporal frequencies. 

Outstanding work on this project during draft PR stage:
* [x] testing as a part of the CESM workflow (requires updates to my CESM version to include CUPiD in the workflow, needs changes to cesm_postprocessing.sh and maybe generate_cupid_config_for_cesm_case.p

#### All PRs Checklist:
* [x] Have you tested your changes as part of the CESM workflow?


mnlevy1981 (changes requested):
I'll wait for you to mark this as "ready for review" before going through it in detail, but one thing that jumped out is a way to avoid needing to add `ninst` to every configuration file.

I didn't really look through the new `config.yml` file, but I noticed it is pointing to output in your scratch space -- I'll make sure you have access to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CESM_output_for_testing/` and then you can copy the files you need over (right now it looks like we're testing on 21 years of ensemble output, maybe we could reduce that to just a few years to save disk space?)

mollymwieringa:
@mnlevy1981 Just a note that the test case I set up definitely shouldn't be processing 21 years-- I think the 3 simulations are each only a month long. I'll double-check, but let me know if that's still too much storage!

mollymwieringa:
@mnlevy1981 Just noting that this is ready for review! I've tested in the CESM workflow, with a couple small changes to the script that sets up the config.yml to get NINST from the CESM case and allow for timeseries production without a baseline case.

PR #406, opened 2026-07-09
Use GenTS for time series workflow

AgentOxygen:
### Description of changes:

Added initial implementation for GenTS time series processing.

#### All PRs Checklist:
It ran on the sample data from the config file, but not all variables were found:

https://github.com/NCAR/CUPiD/blob/392dd8878c1b18868b3d21c086a78c69b185b940/examples/intermediary_outputs/config.yml#L57 

More test cases would be useful.

### DIFF
--- a/cupid/run_timeseries.py
+++ b/cupid/run_timeseries.py
@@ -25,8 +25,6 @@ Options:
 """
 from __future__ import annotations
 
-import os
-import shutil
 import logging
 
 import click
@@ -133,7 +131,7 @@ def run_timeseries(
     for case_index, case_name in enumerate(timeseries_params["case_name"]):
         case_hfc = HFCollection(global_params["CESM_output_dir"] + "/" + case_name, num_processes=num_processes)
         case_hfc = case_hfc.include(inclusive_path_filters).exclude("*/proc/*")
-        
+
         for comp_name in component_run_flags.keys():
             if not component_run_flags[comp_name]:
                 continue
@@ -143,9 +141,9 @@ def run_timeseries(
                 case_hfc = case_hfc.include_years(
                     start_year=comp_config["start_years"][case_index],
                     end_year=comp_config["end_years"][case_index],
-                    glob_patterns=f"*{comp_config["hist_str"]}*"
+                    glob_patterns=f"*{comp_config["hist_str"]}*",
                 )
-        
+
         case_hfc = case_hfc.slice_groups(slice_size_years=5)
         case_tsc = TSCollection(case_hfc, ts_output_dir, num_processes=num_processes)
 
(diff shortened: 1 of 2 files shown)


cupid/run_timeseries.py
```diff
@@ -25,18 +25,20 @@
...
 import click
+from gents.hfcollection import HFCollection
+from gents.timeseries import TSCollection
+from gents.utils import enable_logging
 
 try:
-    import timeseries
     import util
 except ModuleNotFoundError:
-    import cupid.timeseries as timeseries
```
mnlevy1981, line 37: As things stand on `main` right now, `run_timeseries.py` (this script) processes the config file and then passes the proper arguments to `timeseries.create_time_series()`. I can see that being replaced by a call directly into GenTS, but  `timeseries.fix_permissions()` is also a pretty important function and I'm surprised we don't need that anymore. If GenTS can let us set permissions for the output, though, should we remove `timeseries.py` altogether? (We could even move `fix_permissions()` into this script if needed.)

cupid/run_timeseries.py
```diff
@@ -107,130 +94,75 @@ def run_timeseries(
 
     ####################################################################
 
+    enable_logging(verbose=True)
```
mnlevy1981, line 97: Should verbosity be a command-line argument?

cupid/run_timeseries.py
```diff
@@ -80,21 +82,6 @@ def run_timeseries(
...
-        "ice": seaice,
-        "glc": landice,
-        "rof": river_runoff,
-    }
-
-    # Automatically run all if no components specified
-
-    if True not in [atmosphere, ocean, land, seaice, landice, river_runoff]:
-        for key in component_options.keys():
-            component_options[key] = True
```
mnlevy1981, line 96: Did we lose the ability to generate time series for all components simply by omitting any `--comp` flags? I don't see it anywhere. The block above is not very pythonic, but something like

```python3
if not any([atmosphere, ocean, land, seaice, landice, river_runoff]):
    component_run_flags = {key: True for key in component_run_flags}
```

would restore this behavior.

cupid/run_timeseries.py
```diff
@@ -107,130 +94,75 @@ def run_timeseries(
...
+            if not component_run_flags[comp_name]:
+                continue
+            comp_config = timeseries_params[comp_name]
+
+            if "start_years" in comp_config and "end_years" in comp_config:
+                case_hfc = case_hfc.include_years(
+                    start_year=comp_config["start_years"][case_index],
+                    end_year=comp_config["end_years"][case_index],
+                    glob_patterns=f"*{comp_config["hist_str"]}*",
+                )
```
mnlevy1981, line 145: If `start_years` and `end_years` are not provided, will this default to just creating time series over the entire run or will it not produce any time series at all? The former is good, the latter is not.

mnlevy1981 (changes requested):
This first review is just first impressions from looking at the code. I need to clone your branch and test this out, which might lead to additional comments.

cupid/run_timeseries.py
AgentOxygen: Probably best to move `fix_permissions` into the same file so that it remains clear what is does

cupid/run_timeseries.py
AgentOxygen: I made it optional in GenTS and honestly I wish I hadn't. Its pretty helpful for debugging, but if it complicates output then we can make it optional

cupid/run_timeseries.py
AgentOxygen: I will add this back in

cupid/run_timeseries.py
AgentOxygen: It defaults to the full, but it must be a none type or it will throw an error

PR #407, opened 2026-07-22
Allow comparison between more than 2 cases

dsanders02:
### Description of changes:

Rather than case_name and base_case_name we treat case_name as a list and allow for comparison for +2 cases. 

#### All PRs Checklist:
* [x] Have you tested your changes as part of the CESM workflow?


mnlevy1981 (changes requested):
As we talked about in the in-person review, there's a decent amount of clean-up to do:

1. We want to call `consistency_check()` from `get_control_dict()`
2. We want `get_control_dict()` to change some strings into lists of the appropriate length (i.e. `"abc"` -> `["abc", "abc", "abc", "abc"]` when comparing four different cases
3. We want to use python's built-in exceptions instead of click's exceptions
4. There are a few error checks that we can do cleaner (and we can reduce the number of loops while we're at it)
5. There are some files that changed because you brought in modules from other branches for testing, and they should be reverted in this PR

This is really coming along well, though!

mnlevy1981 (approved):
This looks good to me. I think the calls to `plot_diff()` might have the arguments backwards (we want to look at `case[0] - case[n]` for each `n>0`, and I'm pretty sure we're ending up with `case[n] - case[0]` instead) but that seems like the type thing @dabail10 can fix in #412 once this is merged.

@dsanders02 if you think everything is accounted for, can you mark this as "ready to review" and I'll get it merged in.

PR #413, opened 2026-09-10
Initial GenTS implementation

AgentOxygen:
### Description of changes:
* [ ] Please add an explanation of what your changes do and why you'd like us to include them.

#### All PRs Checklist:
* [ ] Have you followed the guidelines in our [Contributor's Guide](https://ncar.github.io/CUPiD/contributors_guide.html)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [ ] Have you made sure that the [`pre-commit` checks passed (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html)?
* [ ] Have you successfully tested your changes locally when running standalone CUPiD?
* [ ] Once you are ready to have your PR reviewed, have you changed it from a Draft PR to an Open PR?

* [ ] Have you [hidden the code cells (#8 in Adding Notebooks Guide)](https://ncar.github.io/CUPiD/addingnotebookstocollection.html) in your notebook?
* [ ] Have you removed any unused parameters from your cell tagged with `parameters`? These can cause confusing warnings that show up as `DAG build with warnings`.
* [ ] Have you moved any observational data that you are using to `/glade/campaign/cesm/development/cross-wg/diagnostic_framework/CUPiD_obs_data` and ensured that it follows this format within that directory: `COMPONENT/analysis_datasets/RESOLUTION/PROCESSED_FIELD_TYPE`?

### DIFF
--- a/cupid/timeseries.py
+++ b/cupid/timeseries.py
@@ -8,21 +8,11 @@ Timeseries generation tool adapted from ADF for general CUPiD use.
 from __future__ import annotations
 
 import glob
-import multiprocessing as mp
 import os
-import subprocess
 from pathlib import Path
 
-import xarray as xr
-
-
-def call_ncrcat(cmd):
-    """This is an internal function to `create_time_series`
-    It just wraps the subprocess.call() function, so it can be
-    used with the multiprocessing Pool that is constructed below.
-    It is declared as global to avoid AttributeError.
-    """
-    return subprocess.run(cmd, shell=False)
+from gents.hfcollection import HFCollection
+from gents.timeseries import TSCollection
 
 
 def fix_permissions(
@@ -77,15 +67,23 @@ def create_time_series(
     """
     Generate time series versions of the history file data. Called by ``cupid-timeseries``.
 
+    Delegates to GenTS, which discovers the history files, reads their time
+    metadata, and writes one file per variable. Output is named
+    ``{case}.{hist_str}.{var}.{YYYYMM-YYYYMM}.nc`` inside ``ts_dir``.
+
     Args
     ----
      - component: str
          name of component, eg 'cam'
-         # This could also be made into a dict and encorporate values such as height_dim
+     - diag_var_list: list
+         variables to create diagnostics (or timeseries) from; the sentinel
+         ``['process_all']`` requests every primary variable in the history files
      - derive_vars: dict
          information on derivable variables
          eg, {'PRECT': ['PRECL','PRECC'],
               'RESTOM': ['FLNT','FSNT']}
+         GenTS performs no computation, so these are derived afterwards by
+         ``derive_cam_variables``; only their constituents are generated here.
      - case_names: list, str
          name of simulaton case
      - hist_str: str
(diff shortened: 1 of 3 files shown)


mnlevy1981:
The following updates let me slice the time series files via `slice_size` in `config.yml`:

```diff
diff --git a/cupid/timeseries.py b/cupid/timeseries.py
index 1d3529c..54de315 100644
--- a/cupid/timeseries.py
+++ b/cupid/timeseries.py
@@ -56,6 +56,7 @@ def create_time_series(
     start_years,
     end_years,
     height_dim,
+    slice_size,
     num_procs,
     serial,
     logger,
@@ -158,6 +159,7 @@ def create_time_series(
             start_years[case_idx],
             end_years[case_idx],
         )
+        hf_collection = hf_collection.slice_groups(slice_size_years=slice_size[case_idx], start_year=start_years[case_idx])
         if len(hf_collection) == 0:
             wmsg = f"WARNING: No {hist_str} history files fall within years"
             wmsg += f" {start_years[case_idx]}-{end_years[case_idx]} for case '{case_name}'."
```

and then the call itself:

```diff
diff --git a/cupid/run_timeseries.py b/cupid/run_timeseries.py
index 5632dec..3280cd3 100755
--- a/cupid/run_timeseries.py
+++ b/cupid/run_timeseries.py
@@ -175,6 +175,13 @@ def run_timeseries(
                     )
             # -----

+            if "slice_size" in timeseries_params[component]:
+                slice_size = timeseries_params[component]["slice_size"]
+                if not isinstance(slice_size, list):
+                    slice_size = len(global_params["case_names"]) * [slice_size]
+            else:
+                slice_size = [(end_year - start_year) + 1 for start_year, end_year in zip(timeseries_params[component]["start_years"], timeseries_params[component]["end_years"])]
+
             # fmt: off
             # pylint: disable=line-too-long
             timeseries.create_time_series(
@@ -190,6 +197,7 @@ def run_timeseries(
                 timeseries_params[component]["start_years"],
                 timeseries_params[component]["end_years"],
                 timeseries_params[component]["level"],
+                slice_size,
                 num_procs,
                 serial,
                 logger,
```

Note that the default is to use a single slice (`size = end_year - start_year + 1`); I was going to update the CUPiD examples to use 10 year chunks, and then I still need to get this plumbed in to the CESM workflow. @AgentOxygen let me know if you want me to push this, or if you're coming up with a better implementation for the `cupid/` directory side of things.

edit: I should probably go with `slice_sizes` for the argument name, since it's a list (and that should be documented somewhere!)

mnlevy1981:
I think what we talked about yesterday was introducing a feature that lets a user say "find the last year of the existing time series files and create new files for each variable starting with the following year" and my suggestion definitely doesn't do that. We were picturing a scenario where a user is running several centuries in 5 year segments, and maybe something like this happens:

1. Run 1 completes 5 years as expected [0001 - 0005]
2. Run 2 is killed in the middle of year 4 (maybe a machine glitch or some other non-reproducible error causes the job to die) [0006 - middle of 0009]
3. User manually starts Run 3 from intermediate restart, it completes 5 years successfully [0009 - 0013]
4. Run 4 completes 5 years as expected [0014 - 0018]
5. etc etc

In this case, CUPiD would run 3 times:

1. After run 1, with history output from 0001 - 0005. Results in time series files for that time period
2. After run 3, with history output from 0001 - 0013. There would already be time series for 0001 - 0005, need 0006 - 0013.
3. After run 4, with history output from 0001 - 0018. Already have time series for 0001 - 0013 (two files per variable), need 0014 - 0018

Doing that without needing to modify any XML variables before runs 3 and 4 to adapt to the unexpected 8-year chunk of output would be awesome. It's just not clear to me if that's something to expect in this PR, or if we want my band-aid solution as a stop-gap measure

mnlevy1981:
I think this PR (plus the changes in https://github.com/AgentOxygen/CUPiD/pull/1) have everything I'd like to make a `v0.7.0` tag to stick in an upcoming CESM tag. There is still an issue of needing to remove the `computed_notebooks` directory so the notebooks run again (#415) but my updates to this PR fix #414 at least

